PR #18, opened 2020-04-10
CAIP-2 chainID

simonovic86:
- convert chainId to CAIP-2 format
- refactor `Context` implementation

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
       "express-validator": "^6.1.0",
       "http-status-codes": "^1.3.2",
       "ipfs-http-client": "^42.0.0",
-      "mail-promise": "^1.2.4",
       "multihashes": "^0.4.19",
       "node-cron": "^2.0.3",
       "reflect-metadata": "^0.1.10",
@@ -47,7 +46,8 @@
       "uuid": "latest",
       "typestub-multibase": "latest",
       "typestub-multihashes": "latest",
-      "cron-parser": "latest"
+      "cron-parser": "latest",
+      "globby": "latest"
    },
    "devDependencies": {
       "@types/chai": "^4.2.11",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -47,7 +46,8 @@
       "uuid": "latest",
       "typestub-multibase": "latest",
       "typestub-multihashes": "latest",
-      "cron-parser": "latest"
+      "cron-parser": "latest",
+      "globby": "latest"
```
oed, line 50: is it really a good idea to have "latest" for the package version? Seems like this could cause problems if there is a release with a breaking change.

src/services/blockchain-service.ts
```diff
@@ -54,6 +59,8 @@ export default class BlockchainService implements Contextual {
 
     const txReceipt: TransactionReceipt = await this.provider.waitForTransaction(txResponse.hash);
     const block: Block = await this.provider.getBlock(txReceipt.blockHash);
-    return new Transaction(txResponse.chainId, txReceipt.transactionHash, txReceipt.blockNumber, block.timestamp);
+
+    const caip2ChainId = 'ethereum:eip155-' + txResponse.chainId;
```
oed, line 63: The correct format should be `'eip155:' + txResponse.chainId`.

package.json
simonovic86: Good point. I'll change versioning.

src/services/blockchain-service.ts
simonovic86: Fixed.

simonovic86:
> This looks good 👍
> 
> Should create a new ticket for making the blockchain service more generalized such that we have a specific ethereum service that conforms to this interface? This would make it easier to add additional blockchains in the future.

Yep. I tried to do that this morning but I didn't like it how it turned out. I'll open a separate ticket for that.

PR #25, opened 2020-04-13
Ethereum blockchain unit tests

simonovic86:
- add more configuration options based on `NODE_ENV` variable
- implement anchor test on local (embedded) ganache instance
- add `husky` for `pre-commit` and `pre-push` git hooks

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
    },
    "husky": {
       "hooks": {
-         "pre-commit": "npm test",
+         "pre-commit": "npm run lint",
          "pre-push": "npm test"
       }
    },
@@ -59,7 +59,6 @@
       "@babel/core": "^7.9.0",
       "@babel/preset-env": "^7.8.4",
       "@babel/preset-typescript": "^7.9.0",
-      "@types/chai": "^4.2.11",
       "@types/cors": "^2.8.6",
       "@types/events": "^3.0.0",
       "@types/express": "^4.17.4",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -59,9 +65,16 @@
       "@types/express": "^4.17.4",
       "@types/jest": "^25.2.1",
       "@types/node": "^13.11.0",
+      "@types/node-cron": "^2.0.3",
       "@types/uuid": "^7.0.2",
+      "@typescript-eslint/eslint-plugin": "^2.19.0",
+      "@typescript-eslint/parser": "^2.19.0",
       "babel-jest": "^25.1.0",
       "chai": "^4.2.0",
```
oed, line 73: I guess this can be removed?

oed: + chai as well?

package.json
```diff
@@ -22,32 +22,38 @@
       "clean": "rm -rf ./build; rm -rf coverage; rm -rf .nyc_output",
       "start": "node ./build/app.js"
    },
+   "husky": {
+      "hooks": {
+         "pre-commit": "npm test",
+         "pre-push": "npm test"
```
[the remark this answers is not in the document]
simonovic86: makes sense 👍 I'll leave just `pre-commit`

package.json
simonovic86: I'll remove those. As for `chai`, I've used its `assert` so I have to replace it with `Jest`s  `expect`.

package.json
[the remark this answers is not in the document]
simonovic86: actually, I left it for `pre-push` and added `npm run lint` for `pre-commit`. We'll have more commits and tests can last longer :)

package.json
simonovic86: removed `chai` as well 👍

PR #38, opened 2020-05-05
Implement CircleCi configuration

simonovic86:
- Implement CircleCi configuration

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,117 @@
+version: 2.1
+
+orbs:
+  aws-ecr: circleci/aws-ecr@6.8.1
+  aws-cli: circleci/aws-cli@1.0.0
+  slack: circleci/slack@3.4.2
+
+jobs:
+  test:
+    working_directory: ~/ceramic-anchor-service
+    docker:
+      - image: circleci/node:10
+    steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,117 @@
+version: 2.1
+
+orbs:
+  aws-ecr: circleci/aws-ecr@6.8.1
+  aws-cli: circleci/aws-cli@1.0.0
+  slack: circleci/slack@3.4.2
```
oed, line 6: Why is this here?

.circleci/config.yml
```diff
@@ -0,0 +1,117 @@
...
+            - push-dev-image
+          filters:
+            branches:
+              only: develop
+      - run-smoke-test:
+          requires:
+            - deploy-dev
+          filters:
+            branches:
+              only: develop
```
oed, line 111: Do we need to run the smoke tests here?

simonovic86: no, I'll remove that

.circleci/config.yml
simonovic86: `slack` doesn't need to be here, I'll remove it :)

PR #119, opened 2020-10-26
Discard not valid requests

simonovic86:
- discard not valid requests
- introduce `DocId` for validation


oed (commented):
Looks like a good start. Now we need to validate the signature in the dag-jose record. Pretty sure we also need to group updates by docid + DID. Lmk if we should sync more on this!

simonovic86:
> Looks like a good start. Now we need to validate the signature in the dag-jose record. Pretty sure we also need to group updates by docid + DID. Lmk if we should sync more on this!

cool, let's sync more about it later 👍

PaulLeCam (approved):

PaulLeCam (approved):

oed (approved):
Looks good on a high level :)

PR #136, opened 2020-11-11
feat(cas): Add API endpoint for fetching the CAIP-2 chainid being used

stbrody:


src/services/blockchain/__tests__/eth-bc-service.test.ts
```diff
@@ -40,22 +40,19 @@ describe('ETH service',  () => {
   });
 
   test('should connect to local ganache server', async () => {
-    try {
-      await ethBc.connect();
-    } catch (err) {
```
stbrody, line 45: @simonovic86 let me know if these try-catch blocks are actually important, but best as I could tell they were just swallowing information about test failures unnecessarily.

stbrody:
I tested this manually using `curl` against the new endpoint, but I didn't see a way to test API endpoint programmatically.  Is it there and I missed it?  If not, I can add a story to our testing epic to add better CAS integration testing, which we could hopefully get to during our Clay hardening phase.  @simonovic86

oed (approved):
Would be nice to have the endpoints documented in the readme.

src/controllers/request-controller.ts
```diff
@@ -117,7 +117,7 @@ export default class RequestController implements Contextual {
   }
 
   @Post()
-  private async create(req: ExpReq, res: ExpRes): Promise<ExpRes<any>> {
+  private async createRequest(req: ExpReq, res: ExpRes): Promise<ExpRes<any>> {
```
simonovic86, line 120: why renaming it to `createRequest`? I guess it's obvious since it's in the `RequestController`.

src/services/anchor-service.ts
```diff
@@ -101,6 +101,10 @@ export default class AnchorService implements Contextual {
     this.requestService = context.lookup('RequestService');
   }
 
+  public async getChainId(): Promise<string> {
```
simonovic86, line 104: Why is this method in the `AnchorService`? You can get the value in the controller from the `BlockchainService`, right?

src/services/blockchain/__tests__/eth-bc-service.test.ts
simonovic86: it's better to remove the `try-catch` 👍

simonovic86 (commented):
@stbrody added a couple of comments

simonovic86:
> I tested this manually using `curl` against the new endpoint, but I didn't see a way to test the new API endpoint programmatically. Is it there and I missed it? If not, I can add a story to our testing epic to add better CAS integration testing, which we could hopefully get to during our Clay hardening phase. @simonovic86

There are no tests for the endpoints still, but feel free to add them 🙂

src/services/anchor-service.ts
stbrody: I guess I would have thought that would be a layering violation?  I had assumed that the BlockchainService would be an implementation detail of the AnchorService.  Currently I don't believe there are any places that access the BlockchainService outside of the AnchorService, so this would be the first.

Anyway I don't feel super strongly about this, if you prefer to get rid of this method and use the blockchain service from `ServiceInfoController` directly I can do that.

stbrody: New version accesses the blockchain service directly

stbrody:
New version up that caches the chainId at startup and errors if it changes over the process lifetime.

simonovic86 (approved):
lgtm

PR #137, opened 2020-11-12
CAS Transactions

simonovic86:
* refactor CAS to use TypeOrm repositories with CLS enabled for transactions
* add test on Ethereum transaction fail and check for requests
    * they will be picked up again upon `expiration` time
* add simple retry mechanism for Ethereum transactions
* use provider from `ethers.js` to estimate Gas price and limit


src/repositories/anchor-repository.ts
```diff
@@ -0,0 +1,39 @@
+import { EntityRepository, InsertResult } from "typeorm";
+import { BaseRepository } from "typeorm-transactional-cls-hooked";
+
+import Context from "../context";
+import Contextual from "../contextual";
+import { Anchor } from "../models/anchor";
+import { Request } from "../models/request";
+
+@EntityRepository(Anchor)
+export default class AnchorRepository extends BaseRepository<Anchor> implements Contextual {
```
stbrody, line 10: Can we add a comment describing what this class is for?  I believe these new repository classes are all about encapsulating the logic for talking to the postgres database that durably stores requests, right?

src/repositories/request-repository.ts
```diff
@@ -0,0 +1,80 @@
...
+import { BaseRepository } from 'typeorm-transactional-cls-hooked';
+
+import Context from "../context";
+import Contextual from "../contextual";
+import { Request, RequestUpdateFields } from "../models/request";
+import { RequestStatus } from "../models/request-status";
+import { config } from "node-config-ts";
+
+@EntityRepository(Request)
+export default class RequestRepository extends BaseRepository<Request> implements Contextual {
```
stbrody, line 12: ditto re: adding a comment.  It's not immediately clear to me what data lives in the RequestRepository and what data lives in the AnchorRepository, so it'd be nice if the comments for these classes could clarify that as well

stbrody (commented):
CAS Transactions is pretty vague as a commit message.  Can you update it to something a bit more descriptive of what is actually changing here?  Is there a github story that this can be linked to that explains the changes?

simonovic86:
> CAS Transactions is pretty vague as a commit message. Can you update it to something a bit more descriptive of what is actually changing here? Is there a github story that this can be linked to that explains the changes?

@stbrody have you read the PRs description, pls? I think it should be pretty clear what changes are there.

src/controllers/internal-controller.ts
```diff
@@ -0,0 +1,47 @@
...
+import Context from '../context';
+import AnchorService from '../services/anchor-service';
+import Contextual from '../contextual';
+import RequestService from '../services/request-service';
+
+import type { Response } from "express-serve-static-core";
+
+@Controller('api/v0/internal')
+@ClassMiddleware([cors()])
+export default class InternalController implements Contextual {
```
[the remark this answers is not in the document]
simonovic86: I removed this controller.

src/repositories/anchor-repository.ts
simonovic86: It should be pretty explanatory what TypeORM repositories are. Here's the [link](https://typeorm.io/#/custom-repository) for more information.

I decoupled database handling logic from the services layer by using repositories. This is not related to the postgres database only since TypeORM provides the same interface for other providers as well.

src/repositories/request-repository.ts
simonovic86: It's pretty straightforward to conclude what data "lives" in repositories by looking at the `@EntityRepository(Entity)` annotation. Here's the [link](https://typeorm.io/#/custom-repository) for more information.

src/context.ts
```diff
@@ -20,21 +21,26 @@ export default class Context {
...
+            // get default exported class
+            // eslint-disable-next-line @typescript-eslint/no-var-requires
+            const clazz = require(absFilename).default;
+            this.register(new clazz());
+          } catch (e) {
+            // do nothing
+          }
         }
+      } else {
+        // @ts-ignore
```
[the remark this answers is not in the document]
simonovic86: I removed it.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -41,42 +43,83 @@ export default class EthereumBlockchainService implements BlockchainService {
     }
```
oed, line 43: At some point we'd likely want to not just use the default provider and explicitly use Infura or similar.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -41,42 +43,83 @@ export default class EthereumBlockchainService implements BlockchainService {
...
-        gasLimit: +config.blockchain.connectors.ethereum.gasLimit,
-        gasPrice: BigNumber.from(config.blockchain.connectors.ethereum.gasPrice),
-      });
+    let { overrideGasConfig } = config.blockchain.connectors.ethereum;
+    if (typeof overrideGasConfig === "string") {
+      overrideGasConfig = overrideGasConfig as string === 'true'
     }
+    if (!overrideGasConfig) {
+      txData.gasPrice = await this.provider.getGasPrice();
+      logger.Info('Estimated Gas price: ' + txData.gasPrice.toString());
```
oed, line 75: This needs to happen inside of the retry loop. Otherwise we could run into trouble when resubmitting and the gasPrice has increased.

oed (changes requested):
mostly just looked at the transaction submitting logic.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
oed: err, see the lines just above this.

src/context.ts
```diff
@@ -20,21 +21,25 @@ export default class Context {
...
-          const clazz = require(absFilename).default;
-          this.register(new clazz());
-        } catch (e) {
-          // do nothing
+  public async build(...paths: (Contextual | string)[]): Promise<void> {
+    for (const elem of paths) {
+      if (typeof elem === 'string') {
+        const filenames: string[] = await Utils.listDir(path.resolve(__dirname, elem));
+        for (const absFilename of filenames) {
+          if (absFilename.endsWith('.map') || absFilename.endsWith('.test.ts')) {
```
PaulLeCam, line 29: If I understand correctly this logic attempts to load files from the filesystem and instantiate classes based on them?
This seems pretty unsafe, especially using this kind of logic to filter files out, for example just adding a `.gitignore` file in one of these folders would break it, no?

src/repositories/anchor-repository.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  setContext(context: Context): void {}
```
PaulLeCam, line 12: It is expected this is a no-op? Seems strange to me to have a class implement an interface with no apparent use for it?

src/repositories/request-repository.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  setContext(context: Context): void {}
+
+  /**
+   * Create/updates client request
+   * @param request - Request
+   */
+  public async createOrUpdate(request: Request): Promise<Request> {
+    return this.manager.getRepository(Request).save(request);
```
PaulLeCam, line 21: I'm not familiar with `typeorm-transactional-cls-hooked` but using a regular TypeORM repository, I think this could be simply `this.save(request)`.

src/repositories/request-repository.ts
```diff
@@ -0,0 +1,80 @@
...
+        })
+      .getMany();
+  }
+
+  /**
+   * Creates new client request
+   * @param cid: Client request CID
+   */
+  public async findByCid(cid: CID): Promise<Request> {
+    return await this.manager.getRepository(Request)
```
PaulLeCam, line 75: `await this.findOne({ cid: cid.toString() })`?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
simonovic86: 👍

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
simonovic86: yep, makes sense 👍

src/context.ts
simonovic86: I guess it will not fail. Pls look at the lines beneath:

```javascript
         try {
            // get default exported class
            // eslint-disable-next-line @typescript-eslint/no-var-requires
            const clazz = require(absFilename).default;
            this.register(new clazz());
          } catch (e) {
            // do nothing
          }
```

src/repositories/anchor-repository.ts
simonovic86: The idea is that only `Contextual` instances will be available in the `Context` for dependency lookup.

src/repositories/request-repository.ts
simonovic86: I'm explicitly getting the `manager` because of it. I guess this is not an issue, right?

src/repositories/request-repository.ts
simonovic86: Same as above.

src/repositories/anchor-repository.ts
stbrody: Maybe I just have a greater preference towards commenting than you do.  Personally I'm of the opinion that just about any class with actual logic in it should have a comment describing what it's responsible for.  If someone's new to this codebase and trying to figure out what this class does, they could see that AnchorRepository extends BaseRepository and implements Contextual, look up Contextual and eventually realize that just means its a singleton owned by the Context object, look up BaseRepository and see that it's imported from the 'typeorm' library, look up the typeorm library, find the documentation, and then realize that this is an ORM class for encapsulating database access logic, presumably for managing the persisted state related to "Anchors".  Or there could be a one sentence comment that just says that and saves them the effort.

I don't believe that someone who's trying to understand this codebase should be required to understand all the libraries that it is built on before they can even begin to start making sense of it.

src/repositories/anchor-repository.ts
stbrody: I think this could be helped by a more descriptive comment for the Contextual interface that clarifies that instances are global singletons managed by the Context object.

src/repositories/request-repository.ts
stbrody: Well, I already spoke above about my feelings on requiring an understanding of all imported libraries in a file in order to understand the code in the file itself, but that actually wasn't what I was even trying to get at with this comment as I had already brought that up above.  Apologies for not being clearer about the difference between I was getting at with this comment vs what I was saying above!  

With this comment I was trying to actually point to ambiguity not around what TypeORM is or is doing, but around what the `Request` and `Anchor` objects represent.  They are ORM Model objects that can be serialized/reconstructed from the database, but what's missing imo is a description of what concept in our code they actually represent.  My understanding is that when the request-controller's 'create' endpoint is activated, a `Request` object is created and persisted into the database.  It then seems that when we periodically perform an anchor, we query for pending Requests from the database and build up `Anchor` objects that in turn get persisted to the database.  It would be good for these basic descriptions of how these objects are used and what they represent to be written down somewhere, presumably in some combination of comments on the `Request` and `Anchor` classes, and comments on the `RequestRepository` and `AnchorRepository` classes.  
Additionally, some other questions that would be nice to answer in such comments would include:
* What's the relationship between a `Request` and an `Anchor`?  Are they one-to-one, or does one `Anchor` record contain information about all the `Requests` that got bundled into a single merkle tree anchor blockchain transaction?
* What's the relationship between an `Anchor` record and the anchor transactions that get created against the blockchain?  Are the `Anchor` records persisted to the database before or after the corresponding transactions are submitted on chain?
* What is the lifetime of both of these records?  Are they garbage collected ever, or do they live in the database forever in a 'complete' state?

Overall I just feel like more comments explaining the workflow, lifecycle, and relationship between the various objects, concepts, and services that are used, implemented, or interacted with in this repository would make it much easier for someone new to get up to speed on this code base.

src/repositories/request-repository.ts
```diff
@@ -0,0 +1,80 @@
...
+    return this.manager.getRepository(Request)
+      .createQueryBuilder()
+      .update(Request)
+      .set(fields)
+      .whereInIds(ids)
+      .execute();
+  }
+
+  /**
+   * Gets all requests by status
```
stbrody, line 52: How about something like:
"Gets all requests in need of processing.  Requests that need processing are Requests that have been received but have never had anchor requests submitted for them, as well as Requests that have been waiting to be successfully anchored for longer than 'expirationPeriod'"

src/repositories/anchor-repository.ts
stbrody: > > CAS Transactions is pretty vague as a commit message. Can you update it to something a bit more descriptive of what is actually changing here? Is there a github story that this can be linked to that explains the changes?
> 
> @stbrody have you read the PRs description, pls? I think it should be pretty clear what changes are there.

Oops, I missed that first comment.  My bad, sorry about that!

stbrody (commented):
Thanks for responding to my initial comments!  Some additional suggestions and clarifications about my previous comments included inline

src/repositories/request-repository.ts
simonovic86: I get what you're saying. Can we open another story (or stories) for CAS comments and documentation since this PR is blocked and it actually solves another set of problems which are really urgent and need to happen in this sprint?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -41,42 +43,85 @@ export default class EthereumBlockchainService implements BlockchainService {
...
-        gasLimit: +config.blockchain.connectors.ethereum.gasLimit,
-        gasPrice: BigNumber.from(config.blockchain.connectors.ethereum.gasPrice),
-      });
+    let { overrideGasConfig } = config.blockchain.connectors.ethereum;
+    if (typeof overrideGasConfig === "string") {
+      overrideGasConfig = overrideGasConfig as string === 'true'
     }
+    if (overrideGasConfig) {
+      txData.gasPrice = BigNumber.from(config.blockchain.connectors.ethereum.gasPrice);
+      logger.Info('Overriding Gas price: ' + txData.gasPrice.toString());
```
oed, line 75: Not sure why you have this if statement both outside and inside of the while loop. I think you should be able to remove it from here.

oed (approved):
Looks good, just one nitpick

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
simonovic86: The first one is just to override it from `config` and to do it just once. The one inside the while loop is if we're not overriding it from `config` but estimating it and it needs to happen on every fail, right?

PR #198, opened 2020-12-02
chore(cas): Cleanup AnchorService

stbrody:
No behavior change here, just refactoring and adding a test


stbrody:
This PR is built on top of Janko's anchor service cleanup PR: https://github.com/ceramicnetwork/ceramic-anchor-service/pull/190

src/controllers/request-controller.ts
```diff
@@ -9,20 +9,20 @@ import cors from 'cors';
...
+import RequestRepository from '../repositories/request-repository';
 
 @singleton()
 @Controller('api/v0/requests')
 @ClassMiddleware([cors()])
 export default class RequestController {
 
   constructor(@inject('anchorService') private anchorService?: AnchorService,
-              @inject('requestService') private requestService?: RequestService, ) {
+              @inject('requestRepository') private requestRepository?: RequestRepository, ) {
```
simonovic86, line 25: I intentionally use `service`s from `controller`s layer. Services can add more logic on top of CRUD repositories. In this case that's not true, but that can change in the future (service layer caching,  etc.).

simonovic86 (approved):
nice cleanup! I added just a comment regarding `controller<->service<->repository` flow.

btw, I needed to merge the `feat/cas-cleanup` and redeploy it on our infra. Feel free to create a PR to develop directly

src/controllers/request-controller.ts
stbrody: Yep, makes total sense, I just figured that's easy enough to add back if/when we need it.  Having the extra indirection layer now with no logic in it just seemed confusing to me.

Issue #199, opened 2020-12-02
Test that anchor proofs can be properly verified against the merkle tree


stbrody:
Ideally this would be a proper integration test between the CAS and js-ceramic to verify that js-ceramic can walk the merkle tree generated by the CAS and verify the anchor proofs.

stbrody:
We now have integration tests between ceramic and the CAS: https://github.com/ceramicnetwork/ceramic-anchor-service/blob/329b52a422763807549be0a2fb4143bffd18a036/src/__tests__/ceramic_integration.test.ts#L321-L354

PR #200, opened 2020-12-02
feat(cas): Use more compact format for merkle tree nodes

stbrody:


src/services/anchor-service.ts
```diff
@@ -53,10 +53,7 @@ class IpfsMerge implements MergeFunction<Candidate> {
   }
 
   async merge(left: Node<Candidate>, right: Node<Candidate>): Promise<Node<Candidate>> {
-    const merged = {
-      L: left.data.cid,
-      R: right.data.cid
-    };
+    const merged = [left.data.cid, right.data.cid];
```
stbrody, line 56: It bothers me that I can change this line without changing the corresponding line below for how we generate the `path` to put into the anchor record, and no tests break.  That's because the `path` is generated here but only consumed in js-ceramic, and we don't have good integration tests between js-ceramic and real CAS nodes yet.  I filed https://github.com/ceramicnetwork/ceramic-anchor-service/issues/199 to fix that.

oed: Thanks for creating that ticket.

oed (approved):
lgtm

simonovic86 (approved):
lgtm!

Issue #172, opened 2020-11-25
Server fails silently when not connected to db


valmack:
I discovered this when deploying a CAS instance to AWS. Security groups on the db were set incorrectly so the instance could not connect to postgres, therefore the server did not _**really**_ start and health checks failed.

However logs were a bit misleading so I didn't realize this

```
Ceramic Anchor Service started in server mode
Ceramic Anchor Service started...
```

### Proposed Solution
- Await connection to db and throw if we fail to connect
- Update logs to be more clear about each step of starting the server

PR #201, opened 2020-12-03
Logger

valmack:
### Motivation
We want to have general logging and logging to support metrics collection.

### Changes

- Replace overnight logger with custom logger
- Create `LOG_LEVEL` env var with options `debug`, `important`, and `warn`
- Wait for db connection. See #172 
- Add http req/resp logging
- Write logs to files if `LOG_TO_FILES` env var (or config) is true


config/default.json
```diff
@@ -30,5 +30,11 @@
         }
       }
     }
+  },
+  "logger": {
+    "level": "debug",
```
stbrody, line 35: I feel like the default should probably be "important" so that 'debug' can have more verbose logs that don't show up by default and have to be opted-into

docker-compose.yml
```diff
@@ -24,6 +24,8 @@ services:
       - ETH_RPC_HOST=http://ganache
       - ETH_RPC_PORT=8545
       - ETH_WALLET_PK=0x06dd0990d19001c57eeea6d32e8fdeee40d3945962caf18c18c3930baa5a6ec9
+      - LOG_LEVEL=debug
```
stbrody, line 27: ditto re: 'important' as the default

src/app.ts
```diff
@@ -153,17 +147,22 @@ export default class CeramicAnchorApp {
     // create connection with database
     // note that it's not active database connection
     // typeorm creates connection pools and uses them for requests
-    createConnection().then(async () => await fn()).catch((e) => {
-      logger.Err(`Failed to start Ceramic Anchor Service. Error ${e.message}`);
+    try {
+        logger.imp('Connecting to database...');
+        const connection = await createConnection();
+        logger.imp(`Connected to database: ${connection.name}`);
+        await fn()
```
stbrody, line 154: Should the call to `fn()` be outside the try/catch block?  Otherwise a failure in the `fn()` call will cause us to log that the database connection failed, even if that's not actually the cause of the failure.

src/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+const LOG_LEVEL = logLevelMapping[config.logger.level] || LogLevel.important
+const LOG_TO_FILES = config.logger.logToFiles || false
+const LOG_PATH = config.logger.filePath || '/usr/local/var/log/cas'
+const METRICS_LOG_PATH = config.logger.metricsFilePath || LOG_PATH + '/metrics'
+
+/**
+ * Handles logging
+ * TODO: Update doc comments
```
stbrody, line 25: TODO?

src/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+export class CASLogger {
+  public readonly logLevel: LogLevel
+  public readonly expressLoggers: any[]
+  private consoleLogger: Logger
+  private includeStackTrace: boolean
+
+  constructor(logLevel: LogLevel = LogLevel.warn) {
```
stbrody, line 33: Here we use 'warn' as the default, on line 18 we use 'important' as the default, and in the config files we use 'debug' is as the default.  Is there a reason for the different defaults at different layers, or should we unify them?

src/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.includeStackTrace = this.logLevel == LogLevel.debug ? true : false
+  }
+
+  // Used for stream interface
+  public write(content: string | object): void {
+    this.info(content)
+  }
+
+  public info(content: string | object): void {
+    if (this.logLevel > LogLevel.debug) return
```
stbrody, line 45: Should this be greater than or equal to?  If the log level is set to debug this should be included right?

src/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.consoleLogger.info(content, this.includeStackTrace)
+  }
+
+  public imp(content: string | object): void {
+    if (this.logLevel > LogLevel.important) return
```
stbrody, line 50: ditto re: greater than or equal to

src/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.logLevel = logLevel
+    this.includeStackTrace = this.logLevel == LogLevel.debug ? true : false
+  }
+
+  // Used for stream interface
+  public write(content: string | object): void {
+    this.info(content)
+  }
+
+  public info(content: string | object): void {
```
stbrody, line 44: We call the log level 'debug' but the method to log at that level is 'info', can we pick one and be consistent?

src/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+  compress: "gzip" // compress rotated files
+})
+
+export const debugLogStream = rfs.createStream(`${LOG_PATH}/debug.log`, {
+  size: "10M", // rotate every 10 MegaBytes written
+  interval: "1d", // rotate daily
+  compress: "gzip" // compress rotated files
+})
+
+export const errorLogStream = rfs.createStream(`${LOG_PATH}/error.log`, {
```
stbrody, line 75: Interesting!  So I guess we get several different log files, one for each level, rather than a single unified log.  I wonder if that'll make it harder to get a unified view from the logs about what is happening at any given moment?  I guess we can always write a simple script to merge the logs based on timestamp if need be.  Do all log lines include the timestamp?

src/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+  size: "10M", // rotate every 10 MegaBytes written
+  interval: "1d", // rotate daily
+  compress: "gzip" // compress rotated files
+})
+
+export const metricsLogStream = rfs.createStream(`${METRICS_LOG_PATH}/out.log`, {
```
stbrody, line 81: Separating the metrics log stream from the diagnostic log stream is great, and addresses my earlier concerns about mixing the two nicely.  Just to make sure I understand, the only messages going into the metrics log file for now will be those generated by the `@ClassMiddleware` annotation on the controller types, and there will be one log message for each request to one of the API endpoints.  Is that right?

src/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+const logLevelMapping = {
+  'debug': LogLevel.debug,
+  'important': LogLevel.important,
+  'warn': LogLevel.warn
+}
+
+const LOG_LEVEL = logLevelMapping[config.logger.level] || LogLevel.important
+const LOG_TO_FILES = config.logger.logToFiles || false
+const LOG_PATH = config.logger.filePath || '/usr/local/var/log/cas'
+const METRICS_LOG_PATH = config.logger.metricsFilePath || LOG_PATH + '/metrics'
```
stbrody, line 21: So, if we're logging api calls to this file, then what we're logging isn't really metrics, but events.  We aggregate those events into metrics, but that happens at a higher level outside of the CAS.  So I think this should probably be call `eventsFilePath`.  In the future, we may want to be able to manually log events here beyond just those auto-generated by calls to the API endpoints.

stbrody (commented):
This is great!  A bunch of little comments and clarifying questions throughout, but I think I'm finally starting to build a mental model of how our logging strategy works thanks to this PR.

config/default.json
valmack: Yeah I was also unclear about what "default" meant and when this json would actually be used.

I guess my idea was we should keep debug on in development?

src/app.ts
valmack: Yup. Good catch

src/logger.ts
valmack: We actually don't need a default here

src/logger.ts
valmack: Yes was trying to keep function name the same so we don't have to change elsewhere.

But using debug works

src/logger.ts
valmack: Yes log lines will include a timestamp. In metrics dashboard we can have a unified view.

src/logger.ts
valmack: No. API req/resp goes into access.log, which the metrics dashboard will read from in addition to the metrics/out.log.

This will be for things other than http api for example low eth.

src/logger.ts
valmack: I think my other comment about access.log vs metrics/out.log clarifies

docker-compose.yml
valmack: Actually if this is in dev mode than i think debug makes sense here

src/logger.ts
valmack: Here if loglevel == important than we __should__ log, not return

config/default.json
stbrody: since I made this comment 5 hours ago I have since come to the conclusion that the exact values of `default.json` don't actually matter, just its structure does as that's used to autogenerate the `Config.d.ts` file which defines the type of the `config` object within the code.  So I think it probably doesn't matter what we put here so long as it's a string.  Would be nice to confirm that though... @oed?

docker-compose.yml
stbrody: Ah, I didn't realize that this is specifically specifying a 'dev' config.  In that case leaving as debug sgtm

src/logger.ts
stbrody: whoops, missed that 'return' in there.  Aaand this is why I'm not a fan of single-line if-statements :sweat_smile:

src/logger.ts
stbrody: okay, there's there's no actual usages currently in this PR, correct?

src/logger.ts
stbrody: Hmm... hard to say without seeing a concrete example, but I suspect even then we wouldn't actually be logging metrics?  Are you imagining log entries would include a field and it's *current* value?  Or will the log messages contain messages that result in a delta to a value that we aggregate on the other end?

stbrody (commented):
All looking good, thanks for answering my questions and correcting my mistaken assumptions!  I think the only open question I have is around whether we should rename the metrics log file to something else.  In my mind if we log things like `{totalFailedRequests: 147}` to that file then it makes sense to call it 'metrics', but if we log things that look like `{event: 'new request', status: 'failed'}` and aggregate it externally, then the log file should be called something else, possibly "events"

config/default.json
stbrody: So I just did some manual testing and I was wrong - the values in default.json do indeed get used as default values if there is no `.env` file present, or for any fields that the `.env` file doesn't specify.

stbrody: We also have example files `config/env/{dev,prod,test}.json` where we can put different recommended defaults for different environments.  But what's set here will be the primary default if nothing else is configured.

valmack: cool so will change to `important`

src/logger.ts
valmack: that's fair

src/logger.ts
valmack: yeah im fine to pull out of this pr and just have it in the next one with metrics logging

src/logger.ts
valmack: not sure yet

simonovic86 (approved):
nice!

PR #204, opened 2020-12-03
IPFS get cache

simonovic86:
- add simple IPFS get cache

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,11 +29,11 @@
       }
    },
    "dependencies": {
-      "@ceramicnetwork/http-client": "^0.8.3",
       "@ceramicnetwork/3id-did-resolver": "^0.5.2",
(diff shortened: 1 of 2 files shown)


stbrody (approved):
Looks great!

Man I love content addressing - I was about to ask about when we have to invalidate the entries, but since they're all content identified we never have to worry about cache invalidation!  Woo!  

https://martinfowler.com/bliki/TwoHardThings.html

simonovic86:
> Looks great!
> 
> Man I love content addressing - I was about to ask about when we have to invalidate the entries, but since they're all content identified we never have to worry about cache invalidation! Woo!
> 
> https://martinfowler.com/bliki/TwoHardThings.html

Oh yeah! Totally agree 🥳

PR #205, opened 2020-12-03
feat(cas): Fix anchor record generation

stbrody:


src/services/anchor-service.ts
```diff
@@ -221,14 +221,14 @@ export default class AnchorService {
...
-   * @param candidates - Merkle tree candidates
    * @param requests - Valid requests
    * @returns An array of Anchor objects that can be persisted in the database with the result
    * of each anchor request.
    * @private
    */
-  async _createAnchorRecords(ipfsProofCid: CID, merkleTree: MerkleTree<Candidate>, candidates: Candidate[], requests: Request[]): Promise<Anchor[]> {
+  async _createAnchorRecords(ipfsProofCid: CID, merkleTree: MerkleTree<Candidate>, requests: Request[]): Promise<Anchor[]> {
     const anchors: Anchor[] = [];
+    const candidates = merkleTree.getLeaves()
```
stbrody, line 231: Super subtle bug.  My original cleanup patch [removed a line](https://github.com/ceramicnetwork/ceramic-anchor-service/commit/8eb5702d25f979553a61a5707754c05e82b9ccd9?branch=8eb5702d25f979553a61a5707754c05e82b9ccd9&diff=unified#diff-2e5303f097304f4422686d5d07e1ff0338a24fa63376be9258cec2c8638754beL154) that re-assigned candidates to the leaves of the merkle tree after generating it.  I had assumed that line was redundant and unnecessary as the leaves of the merkle tree stored the same references to the same candidates as was passed in, but the part I missed was that the merkle tree sorts the candidates.  So we were using an unsorted candidates list here.

I feel like integration tests between js-ceramic and the CAS are probably the highest value tests we can add next sprint...

oed: That and load tests for js-ceramic and CAS.

simonovic86 (approved):
lgtm! 🚀

PR #206, opened 2020-12-03
feat(cas): Enforce depth limit to anchor merkle tree

stbrody:


src/services/anchor-service.ts
```diff
@@ -190,7 +190,19 @@ export default class AnchorService {
...
-      const merkleTree = new MerkleTree<Candidate>(this.ipfsMerge, this.ipfsCompare);
+      if (config.merkleDepthLimit) {
+        const nodeLimit = Math.pow(2, config.merkleDepthLimit)
+        if (candidates.length > nodeLimit) {
+          logger.Warn('Found ' + candidates.length + ' valid candidates to anchor, but our '
+            + 'configured merkle tree depth limit of ' + config.merkleDepthLimit
+            + ' only permits ' + nodeLimit + ' nodes in a single merkle tree anchor proof. '
+            + 'Anchoring the first ' + nodeLimit + ' candidates and leaving the rest for a future anchor batch');
+          candidates = candidates.slice(0, nodeLimit)
+          // TODO: Should we trigger our next anchor faster than the normal interval when there
```
stbrody, line 201: This change takes the approach of just leaving extra requests around to be picked up by the next time we perform an anchor, but doesn't change our anchor schedule in any way.  We may want to consider triggering anchors more frequently if we keep having more requests than fit into a single tree, but I left that as a future improvement for now

config/default.json
```diff
@@ -4,6 +4,7 @@
   "anchorControllerEnabled": false,
   "cronExpression": "0 0 */1 * * *",
   "expirationPeriod": 5400000,
+  "merkleDepthLimit": 10,
```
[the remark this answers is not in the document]
oed: If we anchor once every hour our system would have 0.28 tps with this config :P

src/services/anchor-service.ts
oed: Yeah, I think that is likely the right approach. Anchor as soon as it gets full. We obviously have to watch out for spam though.

stbrody: Sorry, I can't tell if you're agreeing with the way I did it, or the way I describe we could do it :stuck_out_tongue: Do you want me to add a change to anchor early if the number of pending requests gets too large?

config/default.json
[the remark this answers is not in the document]
stbrody: I thought that the values in default.json didn't actually matter and were just used to autogenerate the types in Config.d.ts?  The github story said not to determine what the value should be, so my goal was to leave the default as 'unlimited', which is why I didn't make any change to to the example `.env` files or put a default in the code.  If the values from this file do get used as default values for the config, how do I tell `node-config-ts` to generate a config type with `merkleDepthLimit` as a number, but leave it undefined by default?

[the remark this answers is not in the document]
oed: Ok, that sounds good. I thought this file was setting the default. Probably makes sense to have a default, but we don't need to set it now.

src/services/anchor-service.ts
oed: Sorry, I think I was replying to the code comment directly. I do think triggering the anchor as soon as it get's full makes sense. Is that a large change+

config/default.json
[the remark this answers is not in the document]
stbrody: So I was wrong, it is indeed setting the default.  I couldn't figure out how to make the default be 'undefined' while still assigning the field the type 'number', so I made the default be '0' and interpret zero to mean no limit

src/services/anchor-service.ts
stbrody: It's probably fine if we don't do anything to change the periodic anchor schedule.  If we want to push the periodic anchor back if we did a proactive anchor early, then the change is larger.

stbrody: I take it back, it's not too hard to push back the scheduled anchor too.  Will have a new patch up that does that later today

stbrody: Okay, new version up

stbrody:
ping @oed

config/default.json
```diff
@@ -4,6 +4,7 @@
   "anchorControllerEnabled": false,
   "cronExpression": "0/5 * * * ? *",
   "expirationPeriod": 5400000,
+  "merkleDepthLimit": 0,
```
oed, line 7: I assume we will change this to a reasonable default once we figure out what a reasonable default would be.

oed (approved):
lgtm

PR #231, opened 2020-12-14
feat(cas): Update conflict resolution based on full ceramic document

stbrody:


stbrody:
Still a draft because tests are [failing with a sql foreign key error](https://app.circleci.com/pipelines/github/ceramicnetwork/ceramic-anchor-service/511/workflows/a5b63e5f-832f-4f0d-9373-c38ec19f1a84/jobs/648)

The weird thing is that both of the failing tests can be run on their own and they pass, its only when run after all the other tests in anchor-service.test.ts that they fail.  Commenting out the other tests and running the whole file also works.  So this means that the tests aren't properly isolated and are affecting global state that isn't cleaned up in between tests.  Since the error is a foreign key constraint from the database, my first guess is that the database contents aren't getting cleared between tests, but we do [call DBConnection.clear()](https://github.com/ceramicnetwork/ceramic-anchor-service/blob/01c3e0b69332fbf6154d68be404ab8a82fd43351/src/services/__tests__/anchor-service.test.ts#L115) in between each test, which I thought should take care of that. 

I'm going to try to dig into this a bit more tomorrow, but any advice or ideas on how best to investigate this are welcome! @oed @simonovic86

oed (commented):
Lgtm, not sure what could be up with test though.

stbrody:
New version up that fixes tests by using a transaction to clean up database between tests

oed:
Why are the tests not running on this PR? 🤔

simonovic86 (approved):
lgtm

src/services/ceramic-service.ts
```diff
@@ -28,43 +23,11 @@ export default class CeramicService {
...
-    if (config.ceramic.validateRecords) {
-      this._client = new CeramicClient(config.ceramic.apiUrl);
-
-      const keyDidResolver = KeyDidResolver.getResolver();
-      const threeIdResolver = ThreeIdResolver.getResolver(this._client);
-      this._resolver = new Resolver({
-        ...threeIdResolver, ...keyDidResolver,
-      })
-    }
+    this._client = new CeramicClient(config.ceramic.apiUrl);
```
simonovic86, line 26: The Ceramic service is needed for the `anchor` CAS only and not for the API. If we leave it like this it will create a chicken and egg problem when we need to deploy Ceramic instance and CAS. Ceramic instance needs a CAS api and CAS needs Ceramic API.

simonovic86 (changes requested):
I left one comment related to Ceramic service.

src/services/ceramic-service.ts
oed: Not sure I follow this. Can you elaborate @simonovic86 ?

stbrody: Good call, done

stbrody:
> Why are the tests not running on this PR?

Tests results have never been visible on PRs in ceramic-anchor-service that I've seen.  They do seem to actually be running since when I upload PRs sometimes I get an email from circleCI saying that the tests failed, but I never see the test status on the actual PR page in this repo like I do in js-ceramic.  Would definitely be nice to have that though!

simonovic86 (approved):
lgtm!

PR #232, opened 2020-12-17
Fix IoC issues

simonovic86:


oed (approved):
So this was mostly a typo?

simonovic86:
> So this was mostly a typo?

@oed nope, I just used repository directly and skipped anchor service since anchor service depends on the ceramic service

PR #264, opened 2021-01-12
fix(cas): merkleDepthLimit of zero should mean no limit

stbrody:


src/services/__tests__/anchor-service.test.ts
```diff
@@ -281,6 +281,42 @@ describe('ETH service',  () => {
     expect(requests.length).toEqual(0)
   });
 
+  test('Unlimited anchor requests', async () => {
```
stbrody, line 284: This test passed with no code changes, so this doesn't actually trigger the bug.  I really have no idea why this worked in tests but failed when run for real.  I don't even understand why the code changes I did here are necessary or fixed anything, but I can definitely see the behavior change in manual testing...

stbrody:
I really have no idea why this change is necessary, but it does seem to fix things so... ¯\_(ツ)_/¯

src/merkle/merkle-tree.ts
```diff
@@ -50,7 +50,7 @@ export class MerkleTree<T> {
       throw new Error('Cannot generate Merkle structure with no elements');
     }
 
-    if (this.depthLimit && treeDepth > this.depthLimit) {
+    if (this.depthLimit > 0 && treeDepth > this.depthLimit) {
```
oed, line 53: Any chance that the depthLimit was set to `'0'` (string) or something like that?

oed (approved):

src/merkle/merkle-tree.ts
valmack: Yeah that is likely

src/services/anchor-service.ts
```diff
@@ -101,7 +101,7 @@ export default class AnchorService {
    * @returns whether or not an anchor was performed
    */
   public async anchorIfTooManyPendingRequests(): Promise<boolean> {
-    if (!config.merkleDepthLimit) {
+    if (config.merkleDepthLimit == 0 || config.merkleDepthLimit == undefined) {
```
valmack, line 104: @stbrody I'm still not confident with this. If the merkleDepthLimit is set to '0' what should the behavior here be?

src/services/anchor-service.ts
```diff
@@ -188,7 +188,7 @@ export default class AnchorService {
    */
   async _buildMerkleTree(candidates: Candidate[]): Promise<MerkleTree<Candidate>> {
     try {
-      if (config.merkleDepthLimit) {
+      if (config.merkleDepthLimit > 0) {
```
valmack, line 191: Same thing here. We should probably do explicit type conversion. Or should `config` already have this type set as number?

v-stickykeys (changes requested):

src/merkle/merkle-tree.ts
stbrody: man I hate javascript 😅

stbrody: So... the auto-generated `Config.d.ts` file does have the `merkleDepthLimit` options specified to be of type `number`... but maybe at runtime that isn't actually what happens?

stbrody: sure enough. I logged `typeof config.merkleDepthLimit` and it's a string.  Sigh

src/services/anchor-service.ts
stbrody: well `'0' == 0`, so I think this should be fine?

src/services/anchor-service.ts
stbrody: `config` *does* have this type set as a number, but it seems to be a string at runtime anyway :(
But `'0' > 0 == false`, so I think this is okay?

oed: what if we get `'1'`? is it only a string if 0? Should we parseInt if typeof is string?

stbrody:
Okay, new version up which normalizes the type of the `merkleDepthLimit` field in `config` on process startup

src/services/anchor-service.ts
valmack: didnt realize!

src/app.ts
```diff
@@ -100,11 +100,18 @@ export default class CeramicAnchorApp {
...
-        logger.err(`Unknown application mode ${mode}`);
+        logger.err(`Unknown application mode ${config.mode}`);
         process.exit(1);
       }
     }
-    logger.imp(`Ceramic Anchor Service initiated ${mode} mode`);
+    logger.imp(`Ceramic Anchor Service initiated ${config.mode} mode`);
+  }
+
+  private async _normalizeConfig(): Promise<void> {
```
valmack, line 110: Can we roll in the `_patchConfigTypes` into this?

Otherwise code seems like it could get confusing as the config gets updated in multiple places many times

stbrody: Ooh good call.  Done

v-stickykeys (approved):
LG thanks!

PR #265, opened 2021-01-13
fix: pass full url to ipfs config

### LINKED ISSUE #263: IPFS dag put url broken
On infra seeing requests with a broken url

`FetchError: request to http://https:5011/api/v0/dag/put?format=dag-cbor&input-enc=raw&hash=sha2-256 failed, reason: getaddrinfo ENOTFOUND https`

The env vars we are passing are 

```
IPFS_API_HOST	https://ipfs-clay-cas.3boxlabs.com
IPFS_API_PORT	5011
```

Not clear yet where exactly this is happening.

Screen Shot 2021-01-12 at 2 47 22 PM

valmack:
Closes #263

Due to issue in ipfs-http-client here: https://github.com/ipfs/js-ipfs/issues/3476

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -6,8 +6,7 @@
   "expirationPeriod": 5400000,
   "merkleDepthLimit": 0,
   "ipfsConfig": {
-    "host": "localhost",
-    "port": 5001,
+    "url": "http://localhost:5001",
     "timeout": 120000
   },
   "ceramic": {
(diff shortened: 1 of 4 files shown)


stbrody (approved):

valmack:
I just realized that the type definitions for this config in `config/Config.d.ts` are not in version control, so our postinstall script that runs `node-config-ts` should be run or this could cause errors locally.

From docs: https://www.npmjs.com/package/node-config-ts
> A new Config.d.ts will be generated automatically. This file could be ignored from git as it gets automatically generated based on the structure of default.json

PR #279, opened 2021-01-21
chore(cas): Improve anchor service logging

stbrody:


oed (approved):
lgtm

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -113,6 +113,8 @@ export default class EthereumBlockchainService implements BlockchainService {
           ...txData
         });
         const txResponse: providers.TransactionResponse = await wallet.sendTransaction(txData);
+        // Delete the function to wail until txn is mined so that it doesn't get logged
+        delete txResponse.wait
```
valmack, line 117: Doesn't feel great to delete a method just for this log.

It feels like we should making a call to `txResponse.wait` here

Also seems we can replace `this.provider.waitForTransaction` below with a call to `txResponse.wait`

stbrody: Even if we did actually call the function, that wouldn't make it disappear from the object would it?  You still wind up with the full javascript code of the function printing in the log

valmack (commented):
Just one comment about wait method--trying to avoid future bugs/confusion but not huge

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stbrody: > In any case I'd rather we copy the contents of the response _then_ delete wait.

Yeah, I had just figured that if/when we decide to actually use the function we could copy it out then

valmack: Yes (sorry I just edited my comment)

> In any case I'd rather we copy the contents of the response then delete wait.

stbrody: So I guess there are two options right now.  One is copy the function into a new variable, but then we'll get unused variable warnings.  Or we do a deep clone of the entire `txResponse` object and just delete the `wait` method from the clone.  That's probably the "better" approach, but feels pretty heavyweight considering we don't even use the function right.  Unless you feel strongly about this I'm inclined to leave it as-is and revisit if/when we want to switch to using this `wait` function instead of `provider.waitForTransaction`

valmack: In this case it's not a big deal but in general seems like something we want to avoid. But agree deep clone could be heavy.

Another alternative (which I guess I should have done to begin with) is we just pick out what we want in the log explicitly instead of using spread operator. So everything here except the wait:
https://github.com/ethers-io/ethers.js/blob/master/packages/abstract-provider/src.ts/index.ts#L31-L49

stbrody: Okay that works, done

valmack (approved):
Ty! Looks great.

PR #280, opened 2021-01-22
fix: check undefined config options from rpc

valmack:


src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -30,9 +30,9 @@ export default class EthereumBlockchainService implements BlockchainService {
     const { network } = config.blockchain.connectors.ethereum;
     const { host, port, url } = config.blockchain.connectors.ethereum.rpc;
 
-    if (url != "") {
+    if (url && url != "") {
```
stbrody, line 33: https://memegenerator.net/img/instances/42924369/javascript.jpg

valmack:
The issue here was, when we pass env vars the config option is undefined (so on infra we get an error).

stbrody (approved):
::deep sigh::

PR #285, opened 2021-01-25
Request creation is idempotent

ukstv:
Also present request in a uniform way

### DIFF
--- a/src/controllers/request-controller.ts
+++ b/src/controllers/request-controller.ts
@@ -1,28 +1,32 @@
-import { BAD_REQUEST, CREATED, NOT_FOUND, OK } from 'http-status-codes';
+import { StatusCodes } from 'http-status-codes';
 import { Request as ExpReq, Response as ExpRes } from 'express';
 
 import { config } from 'node-config-ts';
-import awsCronParser from "aws-cron-parser";
+import awsCronParser from 'aws-cron-parser';
 
 import cors from 'cors';
 import { ClassMiddleware, Controller, Get, Post } from '@overnightjs/core';
 
 import CID from 'cids';
 import { RequestStatus } from '../models/request-status';
-import AnchorRepository from "../repositories/anchor-repository";
+import AnchorRepository from '../repositories/anchor-repository';
 import RequestRepository from '../repositories/request-repository';
-import { Anchor } from '../models/anchor';
-import { Request } from "../models/request";
-import { inject, singleton } from "tsyringe";
+import { Request } from '../models/request';
+import { inject, singleton } from 'tsyringe';
 import { logger } from '../logger';
+import { RequestPresentation } from '../models/request-presentation';
 
 @singleton()
 @Controller('api/v0/requests')
 @ClassMiddleware([cors()])
 export default class RequestController {
+  #requestPresentation: RequestPresentation;
 
-  constructor(@inject('anchorRepository') private anchorRepository?: AnchorRepository,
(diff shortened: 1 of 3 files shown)


src/models/request-presentation.ts
```diff
@@ -0,0 +1,69 @@
+import AnchorRepository from '../repositories/anchor-repository';
+import { InvalidRequestStatusError, RequestStatus } from './request-status';
+import awsCronParser from 'aws-cron-parser';
+import { config } from 'node-config-ts';
+import { Request } from './request';
+
+export class RequestPresentation {
```
stbrody, line 7: class comment please

src/models/request-status.ts
```diff
@@ -4,3 +4,9 @@ export enum RequestStatus {
   COMPLETED = 2,
   FAILED = 3,
 }
+
+export class InvalidRequestStatusError extends Error {
```
stbrody, line 8: We have plans to introduce a proper error system based on https://github.com/ceramicnetwork/js-errors-utils.  Until we do that, I'd rather keep our errors simple and hierarchy free, so they're easy to update to the new system later.  Can you just use raw `Error` for now?

src/controllers/request-controller.ts
```diff
@@ -117,46 +66,47 @@ export default class RequestController {
...
+        request.cid = cid.toString();
+        request.docId = docId;
+        request.status = RequestStatus.PENDING;
+        request.message = 'Request is pending.';
+
+        request = await this.requestRepository.createOrUpdate(request);
+
+        const cron = awsCronParser.parse(config.cronExpression);
+
+        return res.status(StatusCodes.CREATED).json({
```
stbrody, line 96: might as well move this case into requestPresentation and use it here, to keep everything in the same place.

stbrody (commented):
a few small comments, but mostly looks good!

src/models/request-presentation.ts
ukstv: ok

src/models/request-status.ts
ukstv: While I agree on unification, I would like to leave it here as it is. Thanks to `never` argument it exhaustively checks if all the enum variants are handled.

Attached for reference: https://github.com/ceramicnetwork/js-errors-utils/issues/2

src/controllers/request-controller.ts
ukstv: whoops, true

src/models/request-presentation.ts
ukstv: hopefully, can be resolved

src/models/request-status.ts
stbrody: wow, fascinating.  What if you removed the `default` case from the switch statement, and put the throw after the switch statement - wouldn't that have the same effect?

stbrody (approved):
I'd still prefer to remove the new Error class, but I'll leave the final decision to you.  Otherwise LGTM

src/models/request-status.ts
ukstv: Not at all. That's the point. With `throw` in `default` case, the compiler does not complain when you add a new variant to enum and forget to handle it.

stbrody: If you have no default case, the compiler won't warn you if you don't handle an enum variant case?  That's pretty disappointing

oed (approved):
lgtm, mod merge requests.

ukstv:
Ooh, that should have been squashed. Sorry for that.

PR #288, opened 2021-01-27
chore(cas): Improve CAS logging

stbrody:


src/controllers/healthcheck-controller.ts
```diff
@@ -26,7 +26,7 @@ export default class HealthcheckController {
 
       return res.status(OK).send()
     } catch (err) {
-      logger.err(err);
+      logger.err(`Performing healtcheck failed: ${err.message()}`);
```
valmack, line 29: `Healtcheck failed: ${err.message()}`

the "performing" is a bit confusing/misleading

src/services/anchor-service.ts
```diff
@@ -305,10 +305,11 @@ export default class AnchorService {
 
     let request = null;
     for (let index = 0; index < requests.length; index++) {
+      let docId
       try {
         request = requests[index];
 
-        const docId = this._getRequestDocID(request)
+        docId = this._getRequestDocID(request)
```
valmack, line 312: Why the change to let here?

src/services/anchor-service.ts
```diff
@@ -317,7 +318,7 @@ export default class AnchorService {
         const candidate = new Candidate(new CID(request.cid), request.id, doc);
         groupedCandidates[candidate.docId] = groupedCandidates[candidate.docId] ? [...groupedCandidates[candidate.docId], candidate] : [candidate];
       } catch (e) {
-        logger.err(e);
+        logger.err(`Error while loading document ${docId.baseID.toString()} at commit ${docId.commit.toString}. Error: ${e.message}`)
```
valmack, line 321: This message breaks if docId is null/undefined. We should probably throw in that case or have an additional try/catch for it above.

valmack (commented):
Awesome! Some small comments

src/services/anchor-service.ts
oed: I guess because `docId` is used in the catch scope.

src/services/anchor-service.ts
oed: Just add some `?`

oed (approved):
lgtm, mod comments.

src/controllers/healthcheck-controller.ts
valmack: Oh sorry @stbrody didn't see full context here. But your latest change makes sense.

src/controllers/anchor-controller.ts
```diff
@@ -1,6 +1,6 @@
 import { INTERNAL_SERVER_ERROR, OK } from 'http-status-codes';
 import { Request as ExpReq, Response as ExpRes } from 'express';
-import { Logger, Logger as logger } from '@overnightjs/logger';
+import { logger } from "../logger";
```
valmack, line 3: Small but would prefer to keep single vs double quotes consistent

valmack (approved):
Great!

src/controllers/anchor-controller.ts
stbrody, line 3:
```suggestion
import { logger } from `../logger`;
```

stbrody: done

PR #292, opened 2021-02-01
chore(cas): Cleanup logger to prepare it to be its own package

stbrody:
Splits `index.ts` into `index.ts` and `logger.ts`.  The idea is that when this moves to be its own package, the `logger.ts` will become the `index.ts` of the new `logger` package.  And then any other packages that want to use the new logger package (the cas, the rest of js-ceramic) will need something like the current `index.ts` to configure and export specific logger instances as needed for the specific package.

I guess that means that the js-ceramic monorepo will need a place to export the specific `logger` instances it needs, someplace other than the `@ceramicnetwork/logger` sub-repo.  I guess I will add a logger.ts file to the `common` package to do that.

### DIFF
--- a/src/logger/index.ts
+++ b/src/logger/index.ts
@@ -1,23 +1,8 @@
-import { Logger, LoggerModes } from '@overnightjs/logger';
-import * as logfmt from 'logfmt';
 import morgan from 'morgan';
 import { config } from 'node-config-ts';
 import path from 'path';
-import util from 'util';
 import { RotatingFileStream } from './stream-helpers';
-
-enum LogStyle {
-  info = 'info',
-  imp = 'imp',
-  warn = 'warn',
-  err = 'err'
-}
-
-enum LogLevel {
-  debug = 1,
-  important = 2,
-  warn = 3
-}
+import { DiagnosticsLogger, LogLevel, ServiceLogger } from './logger';
 
 const logLevelMapping = {
   'debug': LogLevel.debug,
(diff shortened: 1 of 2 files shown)


oed (approved):
> I guess that means that the js-ceramic monorepo will need a place to export the specific logger instances it needs, someplace other than the @ceramicnetwork/logger sub-repo. I guess I will add a logger.ts file to the common package to do that.

Can you expand on this? Why couldn't all loggers be exported from the logger package?

stbrody:
> Can you expand on this? Why couldn't all loggers be exported from the logger package?

The logging types/classes can be, but the logging *instances* can not.  Take a look at the `logEvent` and `logMetric` instances exported in index.ts in this PR.  They contain events and metrics that are specific to the CAS and may not make sense for js-ceramic.  Additionally, the way the loggers get configured may be different.  We don't want to force any package that uses the logger to use the same config file format for exposing logger options.  Also for the CAS I left the default log directory as the CWD, but in js-ceramic I plan to make it default to `~/.ceramic`.

So we need a place in each repo that uses the logger to handle the configuration, and then export specific logger instances based on that configuration

oed:
I see, ok. I just assumed that this would happen top level in the cli. I'm probably missing something here though.

stbrody:
> I see, ok. I just assumed that this would happen top level in the cli. I'm probably missing something here though.

I think we'll expand the `CeramicConfig` interface that is passed to `Ceramic.create()` to include logger-specific arguments, which `Ceramic.create()` will use to call into the `logger.ts` file in the `common` package (via a new `configureLogger()` function I imagine) to control what logger instances it exports.  Then the CLI will be responsible for parsing relevant options out of the config file and passing them to `Ceramic.create`.

oed:
@stbrody so there will be one logger in core, and another one in the http-daemon?

stbrody:
> @stbrody so there will be one logger in core, and another one in the http-daemon?

No, there will just be one logger for all of js-ceramic.  It will be exported from `common`, initialized in core, and configured in the cli

valmack (approved):
LGTM!

PR #298, opened 2021-02-05
chore(cas): Add additional CAS logging

stbrody:


src/services/anchor-service.ts
```diff
@@ -310,10 +318,12 @@ export default class AnchorService {
         request = requests[index];
 
         docId = this._getRequestDocID(request)
+        logger.debug(`About to load ceramic doc ${docId.baseID.toString()} at commit ${docId.commit.toString()}`)
         const doc = await this.ceramicService.loadDocument(docId)
         if (!doc) {
           throw new Error(`No valid ceramic document found with docId ${docId.toString()}`)
         }
+        logger.debug(`Successfully loaded ceramic doc ${docId.baseID.toString()} at commit ${docId.commit.toString()}`)
```
valmack, line 326: Can we include some info about the doc? Maybe the size? Are there other things that are reasonable to log for this specific issue we are seeing now?

stbrody: hard to say.  I guess I could get an approximation of the doc size in there by using DoctypeUtils.serializeState() to get a string representation then taking the length of that string.  @oed any thoughts here?

valmack (approved):
Given the hanging is not an issue maybe we don't need to be this verbose. Will leave that to you to decide if there is clean up, or things that can be moved down to debug level.

I think additional things we may want to log (e.g. doc info) don't necessarily need to be in this PR if we want to get this in quickly. Again, will leave to you to decide.

stbrody:
Okay cool, I removed the messages before and after each individual document load as I think that will be pretty verbose.  But I left all the other logs that announce when we move between phases of the anchor process

Issue #295, opened 2021-02-03
Get forwarded request ip


valmack:
Need to enable `trust proxy` to log the ip as the x-forwarded-for address instead of load balancer address

See: https://expressjs.com/en/guide/behind-proxies.html

Can also consider doing this in js-ceramic

PR #299, opened 2021-02-07
feat: get forwarded ip in access logs

valmack:
Related to #295 and allows us to see requests per origin properly in metrics

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -19,6 +19,7 @@ export default class CeramicAnchorServer extends Server {
   constructor(private container: DependencyContainer) {
     super(true);
 
+    this.app.set('trust proxy', true);
     this.app.use(bodyParser.json());
     this.app.use(bodyParser.urlencoded({ extended: true }));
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -19,6 +19,7 @@ export default class CeramicAnchorServer extends Server {
   constructor(private container: DependencyContainer) {
     super(true);
 
+    this.app.set('trust proxy', true);
```
stbrody, line 22: wow, magic!  I'm just going to trust you that this is the proper magic 😅

stbrody (approved):

stbrody:
I assume there needs to be a follow-up to make this actually show up as a field in the event logs?

src/server.ts
valmack: > I assume there needs to be a follow-up to make this actually show up as a field in the event logs?

No we are already capturing the proper field `ip`. This change sets that value to the X-Forwarded-For ip instead of the load balancer's ip.

PR #300, opened 2021-02-08
ci(fix): temporarily remove deploy step

valmack:
CI is failing because the cluster it is trying to deploy to no longer exists


oed (approved):
The goal here is that it should auto deploy to dev-unstable?

valmack:
@oed When we next update this develop will deploy to dev and main to clay. For now this just prevents the red X from ci failing

PR #314, opened 2021-03-02
feat(cas): Add merkle tree metadata with bloom filter

stbrody:


package.json
```diff
@@ -39,6 +39,7 @@
       "@overnightjs/core": "^1.7.5",
       "aws-cron-parser": "1.1.5",
       "base64url": "^3.0.1",
+      "bloomfilter.js": "^1.0.0",
```
stbrody, line 42: Went with [this one](https://www.npmjs.com/package/bloomfilter.js) mostly because it's MIT license and had serialization functionality.  Didn't do much however in terms of verifying that the implementation is solid.  Not sure if there's any good way to get signal on that.

src/merkle/merkle-objects.ts
```diff
@@ -0,0 +1,94 @@
+import CID from "cids";
```
stbrody, line 1: open to ideas for a better name for this file

src/merkle/merkle-objects.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Implements IPFS merge CIDs
+ */
+export class BloomMetadata implements MetadataFunction<Candidate, TreeMetadata> {
+  generateMetadata(leaves: Array<Node<Candidate>>): TreeMetadata {
+    const bloomFilterEntries = new Set<string>()
+    for (const node of leaves) {
+      const doc = node.data.document
+      bloomFilterEntries.add(`docid-${doc.id.baseID.toString()}`)
+      bloomFilterEntries.add(`schema-${doc.metadata.schema?.toString()}`)
+      bloomFilterEntries.add(`family-${doc.metadata.family}`) // todo should we use a special charactor of some kind to distingush the family "undefined" from no family?
```
stbrody, line 77: Currently if no schema or family is set, they still go into the bloom filter as `schema-undefined` or `family-undefined`.  Putting *something* in when nothing is provided makes sense to me cause I could imagine indexing services caring about docs with no schema defined at all, or no family defined at all.  In the case of schema, `schema-undefined` is fine to be the marker for "no schema was set" since "undefined" isn't a valid `CommitID` and thus can never be the actual value of the schema field.  For "family", however, someone could in theory set their "family" to the string "undefined", and we wouldn't be able to distinguish that from no family being set at all.  The only way I could imagine getting around that would be to have some special character that we forbid being present in the "family" field

stbrody: Actually nevermind, we can change the part that is hard-coded (`family-`) in a way that no user will be able to affect

src/merkle/merkle-objects.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (doc.metadata.tags) { // TODO should we add a marker for empty tags?
```
stbrody, line 78: While it makes sense to include markers for no family and no schema, I was less sure about something for no tags, especially since there can be multiple but the bloom filter doesn't give any other indication of how many tags there are per document.  So I'm leaning towards leaving tags as-is and just not putting any tags-related entries in for documents with no tags

src/merkle/merkle-objects.ts
stbrody: changed to `no-family` and `no-schema` as the markers in the bloom filter for when no family or no schema is provided

package.json
oed: What were the alternatives you found? Iirc last time I looked there wasn't that many around. 
Maybe we should just use the eth bloom filter? https://github.com/hyperdivision/eth-bloomfilter
That should be well documented and implemented in other langs. Codebase is minimal and maintained by a really solid dev.

src/merkle/merkle-objects.ts
oed: Including `no-schema` and `no-schema` makes no sense to me. Why would you want that?

oed (changes requested):

src/merkle/merkle-objects.ts
oed: Ok, I see your comment above. To me it still doesn't make sense to include them because no one would want to index for "all documents without a schema/family".

package.json
stbrody: Looking at the code I see two problems with using that one.  The first more minor issue is that the size of the bloom filter is fixed at 256 bits, it doesn't scale with the size of the input or the desired false positive rate.  The bigger issue is that it doesn't have any serialization/deserialization functionality, which we need to persist the filter into the `TreeMetadata` ipfs record.

oed: **Serialization** it uses a buffer which can easily be serialized/deserialized

**Size** I guess we should figure out if 256 *bytes* is actually a problem?

stbrody: But the internal buffer isn't exposed.  Are you suggesting we make a PR to that library to allow constructing the bloom filter from a given buffer?

oed: yes it's exposed on a public property `.buffer`.

src/merkle/merkle-objects.ts
stbrody: Interesting.  I agree no one is likely to index *just* on "all documents with no schema".  I could, however, imagine a use case for something like "index all docs with this controller and this family but no schema", or "index all docs with this controller and this tag but no family".  That said I don't feel super strongly about supporting such a use case, so I'm fine with leaving it out for now until/unless someone asks for it down the road

package.json
stbrody: New version up that switches back to the originally proposed library: https://www.npmjs.com/package/bloom-filters.
Suppressing the compilation warning worked trivially

src/merkle/merkle-objects.ts
```diff
@@ -0,0 +1,98 @@
...
+  TreeMetadata,
+} from './merkle';
+
+import { logger } from '../logger';
+import { IpfsService } from '../services/ipfs-service';
+
+import { BloomFilter } from 'bloom-filters';
+
+const BLOOM_FILTER_TYPE = "jsnpm_bloom-filters";
+const BLOOM_FILTER_FALSE_POSITIVE_RATE = 0.0001
```
stbrody, line 17: Just FYI:
```
# node
Welcome to Node.js v14.15.5.
Type ".help" for more information.
> const { BloomFilter } = require('bloom-filters')
undefined
> const NUM_ITEMS = 10000
undefined
> const FALSE_POSITIVE_RATE = 0.0001
undefined
> BloomFilter.create(NUM_ITEMS, FALSE_POSITIVE_RATE).size
191702

```

src/merkle/merkle-objects.ts
oed: > "index all docs with this controller and this family but no schema"

Adding `no-schema` wouldn't really help with this use case since you can't know if this was added for the doucment which you care about or any other documents.

src/merkle/merkle-objects.ts
oed: Isn't this an extremely low false positive rate?

oed (commented):
Looks good overall

src/merkle/merkle-objects.ts
stbrody: Well, the only cost to a lower false positive rate is a greater size of the resulting filter.  From my simple experiments, it seemed that the number of entries had a bigger impact on the filter size than the false positive rate:

```
> BloomFilter.create(100, 0.0001).size
1918
> BloomFilter.create(1000, 0.0001).size
19171
> BloomFilter.create(10000, 0.0001).size
191702
> BloomFilter.create(100000, 0.0001).size
1917012
> BloomFilter.create(10000, 0.01).size
95851
> BloomFilter.create(10000, 0.001).size
143776
> BloomFilter.create(10000, 0.0001).size
191702
> BloomFilter.create(10000, 0.00001).size
239627
```

stbrody: We could go down to `.001`, but it looks like a 10x increase in false positives only gains us ~25% space usage improvement

oed: Ok, thanks for the rundown. I'm not sure what the best call is here actually. I guess we can start with your configuration and then see if any issues arise on the indexing side.

stbrody: Care to approve the PR?

oed (approved):

PR #317, opened 2021-03-04
fix(cas): Add timeout to how long we wait for eth transactions to get mined

stbrody:
This change increases the rate at which the ethers library will poll infura from every 4 seconds to every 15 seconds (just over the block production rate), and adds a 1 hour timeout to how long we wait for each submitted transaction to be mined.

There is one failure mode I'd like help thinking through whether or not this is a security concern and if there's anything we can do to mitigate it.  What if we submit a transaction that doesn't get mined (let's say because gas price is too low).  We time it out after an hour and retry up to 3 times (recalculating gas price each time).  But after the final retry times out, we declare the transaction failed and move on.  The anchor batch is declared a failure, and the pending CIDs to be anchored stick around in the database and get picked up in the next anchor batch (which we'll assume is successful).

What if, after we give up on the transaction after an hour (and with 3 retries), the transaction actually *does* wind up eventually getting mined (say, because gas prices come down).  Now the CIDs were anchored twice, but only the second time they were anchored generated an anchor proof and corresponding ceramic anchor commits.  Now a bunch of time passes and new updates are made on top of some of the documents that were updated.  What if someone inspects the blockchain and realizes that there was an anchor batch that made it on chain, but never had AnchorCommits published for them?  They could then go and publish AnchorCommits from this batch that actually succeeded but we gave up on, and in so doing revert the history on a bunch of documents from that batch.

The problem is, I'm not sure there's anything we can actually do to prevent this with our current anchoring model.  We could wait longer before giving up on a transaction, but there's never a time after which we can be 100% certain it won't be mined in the future.  Am I missing something @oed?

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -26,6 +26,7 @@
           "port": "8545",
           "url": ""
         },
+        "transactionTimeoutSecs": 60,
         "account": {
           "privateKey": "0x06dd0990d19001c57eeea6d32e8fdeee40d3945962caf18c18c3930baa5a6ec9"
         }
--- a/config/env/dev.json
+++ b/config/env/dev.json
@@ -26,6 +26,7 @@
           "port": "@@ETH_RPC_PORT",
           "url": "@@ETH_RPC_URL"
         },
+        "tranactionTimeoutSecs": "@@ETH_TXN_TIMEOUT",
         "account": {
           "privateKey": "@@ETH_WALLET_PK"
         }
--- a/src/services/blockchain/__tests__/eth-bc-service.test.ts
+++ b/src/services/blockchain/__tests__/eth-bc-service.test.ts
@@ -16,6 +16,7 @@ let ethBc: BlockchainService = null;
 
 describe('ETH service',  () => {
   jest.setTimeout(25000);
+  const blockchainStartTime = new Date(1586784002000)
   beforeAll(async () => {
     container.register("blockchainService", {
       useClass: EthereumBlockchainService
@@ -25,7 +26,7 @@ describe('ETH service',  () => {
 
     ganacheServer = Ganache.server({
       gasLimit: 7000000,
-      time: new Date(1586784002855),
+      time: blockchainStartTime,
       mnemonic: 'move sense much taxi wave hurry recall stairs thank brother nut woman',
       default_balance_ether: 100,
       debug: true,
@@ -55,6 +56,15 @@ describe('ETH service',  () => {
     const cid = new CID('bafyreic5p7grucmzx363ayxgoywb6d4qf5zjxgbqjixpkokbf5jtmdj5ni');
     const tx = await ethBc.sendTransaction(cid);
     expect(tx).toBeDefined();
+
+    // checking the timestamp against the snapshot is too brittle since if the test runs slowly it
+    // can be off slightly.  So we test it manually here instead.
+    const blockTimestamp = tx.blockTimestamp
+    delete tx.blockTimestamp
+    const startTimeSeconds = blockchainStartTime.getTime() / 1000
+    expect(blockTimestamp).toBeGreaterThan(startTimeSeconds)
+    expect(blockTimestamp).toBeLessThan(startTimeSeconds + 5)
+
     expect(tx).toMatchSnapshot();
   });
 
--- a/src/services/blockchain/__tests__/eth-bc-service.test.ts
+++ b/src/services/blockchain/__tests__/eth-bc-service.test.ts
@@ -61,7 +61,7 @@ describe('ETH service',  () => {
     // can be off slightly.  So we test it manually here instead.
     const blockTimestamp = tx.blockTimestamp
     delete tx.blockTimestamp
-    const startTimeSeconds = blockchainStartTime.getTime() / 1000
+    const startTimeSeconds = Math.floor(blockchainStartTime.getTime() / 1000)
     expect(blockTimestamp).toBeGreaterThan(startTimeSeconds)
     expect(blockTimestamp).toBeLessThan(startTimeSeconds + 5)
 
--- a/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
+++ b/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
@@ -10,6 +10,7 @@ import { logger, logEvent, logMetric } from "../../../logger";
 import Transaction from "../../../models/transaction";
 import BlockchainService from "../blockchain-service";
 import { TransactionRequest } from "@ethersproject/abstract-provider";
+import Utils from '../../../utils';
 
 const BASE_CHAIN_ID = "eip155";
 const TX_FAILURE = 0;
@@ -202,7 +203,7 @@ export default class EthereumBlockchainService implements BlockchainService {
           throw new Error("Failed to send transaction");
         } else {
           logger.warn(`Failed to send transaction; ${MAX_RETRIES - attemptNum} retries remain`)
-          await new Promise(resolve => setTimeout(resolve, 5000));
+          await Utils.delay(5000)
         }
       }
     }
--- a/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
+++ b/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
@@ -119,6 +119,7 @@ export default class EthereumBlockchainService implements BlockchainService {
       data: hexEncoded,
       nonce: baseNonce,
     };
+    const transactionTimeoutSecs = config.blockchain.connectors.ethereum.transactionTimeoutSecs
 
     let attemptNum = 0;
     while (attemptNum < MAX_RETRIES) {
@@ -149,7 +150,7 @@ export default class EthereumBlockchainService implements BlockchainService {
         }
 
         const txReceipt: providers.TransactionReceipt = await this.provider.waitForTransaction(
-          txResponse.hash, 1, config.blockchain.connectors.ethereum.transactionTimeoutSecs * 1000);
+          txResponse.hash, 1, transactionTimeoutSecs * 1000);
         logEvent.ethereum({
           type: 'txReceipt',
           ...txReceipt
@@ -186,6 +187,13 @@ export default class EthereumBlockchainService implements BlockchainService {
               logger.err(errMsg);
               throw new Error(errMsg);
             }
+          } else if (code === ErrorCode.TIMEOUT) {
+            logEvent.ethereum({
+              type: 'transactionTimeout',
+              transactionTimeoutSecs,
+            });
+            logger.err(`Transaction timed out after ${transactionTimeoutSecs} seconds without being mined`);
+            // Fall through and retry if we have retries remaining
           }
         }
 
--- a/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
+++ b/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
@@ -14,6 +14,7 @@ import { TransactionRequest } from "@ethersproject/abstract-provider";
 const BASE_CHAIN_ID = "eip155";
 const TX_FAILURE = 0;
 const TX_SUCCESS = 1;
+const MAX_RETRIES = 3;
 
 const POLLING_INTERVAL = 15 * 1000 // every 15 seconds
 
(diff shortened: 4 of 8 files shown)


stbrody:
I guess one thing we could do is maintain a copy of our wallet's `nonce` in the local database, and every time we go to submit a transaction to the blockchain confirm that the nonce we're about to use is the same as what we know about from our local database.  Then if we see that the nonce on chain is higher than the nonce we have stored locally we could... fetch every transaction since the last nonce we were aware of, and try to generate AnchorProofs/AnchorCommits for every CID that was actually anchored?  But by that point the ceramic node(s) that submitted the anchor requests might have moved on and they're what normally creates the AnchorProofs...

oed:
The problem you are describing is not actually a problem afaict. If tx1 doesn't go though, we discard it and make a new tx2 then both txs would have the same nonce. If tx2 ends up getting mined this invalidates tx1.

It seems reasonable still to maybe keep the data about previously failed tx around so that we can check them if there is some other edge case here.

src/services/blockchain/__tests__/eth-bc-service.test.ts
```diff
@@ -55,6 +56,15 @@ describe('ETH service',  () => {
     const cid = new CID('bafyreic5p7grucmzx363ayxgoywb6d4qf5zjxgbqjixpkokbf5jtmdj5ni');
     const tx = await ethBc.sendTransaction(cid);
     expect(tx).toBeDefined();
+
+    // checking the timestamp against the snapshot is too brittle since if the test runs slowly it
+    // can be off slightly.  So we test it manually here instead.
+    const blockTimestamp = tx.blockTimestamp
+    delete tx.blockTimestamp
+    const startTimeSeconds = blockchainStartTime.getTime() / 1000
```
oed, line 64:
```suggestion
    const startTimeSeconds = Math.floor(blockchainStartTime.getTime() / 1000)
```
???

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -38,6 +41,8 @@ export default class EthereumBlockchainService implements BlockchainService {
       this.provider = ethers.getDefaultProvider(network);
     }
 
+    this.provider.pollingInterval = POLLING_INTERVAL
```
oed, line 44: Is this interval polling as in retrying to send the tx, or to poll the tx status, or both?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -131,7 +136,7 @@ export default class EthereumBlockchainService implements BlockchainService {
           throw new Error("Chain ID of connected blockchain changed from " + this.chainId + " to " + caip2ChainId)
         }
 
-        const txReceipt: providers.TransactionReceipt = await this.provider.waitForTransaction(txResponse.hash);
+        const txReceipt: providers.TransactionReceipt = await this.provider.waitForTransaction(txResponse.hash, 1, TRANSACTION_TIMEOUT);
```
oed, line 139: I guess TRANSACTION_TIMEOUT should be scaled with how frequently we run the anchor service. On mainnet I think it would be fine if this is up to 6h.

src/services/blockchain/__tests__/eth-bc-service.test.ts
stbrody: not necessary in this case since the start time is set by the test to an even multiple of 1000, but also doesn't hurt

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stbrody: I'll make it configurable

stbrody:
We should probably discuss the double-anchoring problem live.  I still think there's a case where it could be a problem, although the window is narrow (depending on how soon we retry after a failed batch)

valmack:
@stbrody So as far as nonces we don't have to handle those manually. Ethers handles for us.

In terms of failing a request, we should not be doing this after the time out. In this case of "it didn't get mined within 1 hour" that doesn't mean we fail the batch. It should essentially be infinite retries. If we consistently don't get transacations mined we should be halting the process because that means there is something wrong with our gas estimation or something else in our code is broken.

In other words, there is no point in trying to anchor a new batch if 3 transactions in 3 hours don't get mined. We should retry the same batch maybe with a longer delay.

valmack:
As far as anchor the same batch twice.. I'd be interested to know why that is actually a problem besides costing us $.

stbrody:
> In this case of "it didn't get mined within 1 hour" that doesn't mean we fail the batch. It should essentially be infinite retries.

That's a good point and an interesting idea, but would require rethinking how anchors get scheduled.  Right now if one anchor batch runs too long, a new one will be started and overlap with it.  We should probably either not start a new anchor service instance if an existing one is running, or change to have a single anchor service that is always up and running but just runs batches periodically.

stbrody:
> As far as anchoring the same batch twice.. I'd be interested to know why that is actually a problem besides costing us $.

The problem is if the later anchored one is the one that gets back to the ceramic nodes that requested the anchors, as then they will create and publish AnchorCommits pointing to the later-anchored transaction.  Then someone else malicious who notices that there was an earlier-anchored batch that never had AnchorCommits created and published for it could go and publish those AnchorCommits, reverting the document history for a bunch of documents all at once.

valmack:
> earlier-anchored batch that never had AnchorCommits created and published for it could go and publish those AnchorCommits, reverting the document history for a bunch of documents all at once.

Hmm ok so basically the problem is like a single document diverges
```
 -- A anchored ------------------------------------- A published -- C anchored -- C published -->
 ---- A' anchored  -- A' published -- B anchored -- B published -->
```

So, A and C are valid because A was the earlier anchor
A' becomes invalid once A is published because earliest anchor wins
B becomes invalid once A is published because it points to A'

valmack:
> > In this case of "it didn't get mined within 1 hour" that doesn't mean we fail the batch. It should essentially be infinite retries.
> 
> That's a good point and an interesting idea, but would require rethinking how anchors get scheduled. Right now if one anchor batch runs too long, a new one will be started and overlap with it. We should probably either not start a new anchor service instance if an existing one is running, or change to have a single anchor service that is always up and running but just runs batches periodically.

Actually yeah I think I am wrong about nonces in this case because if we have overlapping processes then the nonce for the new process would be the same as the old one and the tx on the new one would fail if the old one goes through... Not the end of the world but a waste of effort if we know we are waiting on the tx in another process.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stbrody: It's used internally by `@ethersproject`'s Providers to control the frequency of its poll loop, which can be used for any number of registered tasks.  I think in our case however it's primarily how often the `this.provider.waitForTransaction()` call polls internally for the txn to get mined

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -52,6 +57,34 @@ export default class EthereumBlockchainService implements BlockchainService {
...
+    if (config.blockchain.connectors.ethereum.overrideGasConfig) {
+      txData.gasPrice = BigNumber.from(config.blockchain.connectors.ethereum.gasPrice);
+      logger.debug('Overriding Gas price: ' + txData.gasPrice.toString());
+
+      txData.gasLimit = BigNumber.from(config.blockchain.connectors.ethereum.gasLimit);
+      logger.debug('Overriding Gas limit: ' + txData.gasLimit.toString());
+    } else {
+      const gasPriceEstimate = await this.provider.getGasPrice();
+      // Add 10% extra to gas price for each subsequent attempt
+      txData.gasPrice = gasPriceEstimate.add(gasPriceEstimate.mul(attempt * .01));
```
valmack, line 78: Related to comment below. This makes sense to bump gas price if the problem is actually gas.

config/default.json
```diff
@@ -26,6 +26,7 @@
           "port": "8545",
           "url": ""
         },
+        "transactionTimeoutSecs": 60,
```
valmack, line 29: Yup makes sense!

config/env/dev.json
```diff
@@ -26,6 +26,7 @@
           "port": "@@ETH_RPC_PORT",
           "url": "@@ETH_RPC_URL"
         },
+        "tranactionTimeoutSecs": "@@ETH_TXN_TIMEOUT",
```
valmack, line 29: Note to self mostly, 60s with 3 retries is too long if we are anchoring every 2 minutes on dev. We'll run into a nonce issue if we timeout.

valmack (commented):
Just some clarifying questions.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -87,26 +120,10 @@ export default class EthereumBlockchainService implements BlockchainService {
...
-      logger.debug('Overriding Gas price: ' + txData.gasPrice.toString());
-
-      txData.gasLimit = BigNumber.from(config.blockchain.connectors.ethereum.gasLimit);
-      logger.debug('Overriding Gas limit: ' + txData.gasLimit.toString());
-    }
-
-    let retryTimes = 3;
-    while (retryTimes > 0) {
+    let attemptNum = 0;
+    while (attemptNum < MAX_RETRIES) {
```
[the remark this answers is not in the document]
stbrody: Not totally clear to me what the right call here is.  I don't really have enough ethereum experience to say whether or not there are other failure modes that might be recoverable with a retry.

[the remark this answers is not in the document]
stbrody: The fact that even if we fail the whole batch we would still just start a new batch shortly thereafter definitely complicates the error handling story here.  I'm starting to think more and more that we should be handling scheduling within a long-running CAS instead of starting a new instance every batch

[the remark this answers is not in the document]
valmack: > The fact that even if we fail the whole batch we would still just start a new batch shortly thereafter definitely complicates the error handling story here. I'm starting to think more and more that we should be handling scheduling within a long-running CAS instead of starting a new instance every batch

Maybe we made an optimistic optimization here without knowing what it would look like ultimately. Though I'll add that if it is a problem of how we are running the process we can go down the route of just a better way to do that. As an example, add a check to the docker image to see if we already have an ECS task for the anchor service running and if so don't spin up another one.

So we can evaluate where to place the solution based on whatever is easiest + simplifies this.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -171,11 +189,11 @@ export default class EthereumBlockchainService implements BlockchainService {
...
 
-        retryTimes--;
-        if (retryTimes === 0) {
+        attemptNum++;
+        if (attemptNum >= MAX_RETRIES) {
           throw new Error("Failed to send transaction");
         } else {
-          logger.warn(`Failed to send transaction; ${retryTimes} retries remain`)
+          logger.warn(`Failed to send transaction; ${MAX_RETRIES - attemptNum} retries remain`)
           await new Promise(resolve => setTimeout(resolve, 5000));
```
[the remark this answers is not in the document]
valmack: Gotcha. Can we add a little comment here? Just to avoid further questions like mine in the future.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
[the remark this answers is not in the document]
oed: > Do we risk losing eth here?

Not really the only thing that could happen is that our tx doesn't get mined because the `gasPrice` is to low.

The `gasLimit` should always be static for the type of transaction we are making.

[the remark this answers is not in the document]
valmack: > > Do we risk losing eth here?
> 
> Not really the only thing that could happen is that our tx doesn't get mined because the `gasPrice` is to low.
> 
> The `gasLimit` should always be static for the type of transaction we are making.

@stbrody this is kind of the point that I want to be more clear. Since this is not a smart contract tx it really is fine to do it the way we are because there are really only 2 failure modes--one where increasing gas price helps and the other where increasing gas price doesnt do anything. The issue I see in the way this code is written is that the knowledge of this is implicit.

Maybe an easy way to refactor this is

```
while(retrying) {
  try sendSimpleTx & break
  catch timeout & increase gasPrice
}
```

But up to you if it's that important rn.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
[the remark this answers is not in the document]
stbrody: Added a `Utils.delay()` method and use that here, hopefully that makes this clearer

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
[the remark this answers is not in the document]
stbrody: I don't actually follow what your suggestion here is?  That looks basically like how this code is already structured to me.

oed (approved):
When conflict is fixed ;)

PR #318, opened 2021-03-05
ci: remove image build and deploy

valmack:
Using dockerhub instead for now

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,5 @@
 version: 2.1
 
-orbs:
-  aws-ecr: circleci/aws-ecr@6.8.1
-  aws-cli: circleci/aws-cli@1.0.0
-
 jobs:
   test:
     working_directory: ~/cas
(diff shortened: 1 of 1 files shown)


oed:
Sorry I don't have the full context here. What are these being replaced with?

valmack:
@oed docker hub images here: https://hub.docker.com/repository/docker/ceramicnetwork/ceramic-anchor-service

stbrody:
I also don't really feel like I have enough context to meaningfully sign off on this.  Maybe @zachferland does, or maybe no one besides you really does?  In which case I guess you should just push without review?  Hopefully soon we'll hire a backend engineer and have another person on the team who's familiar with how our infra is managed...

valmack:
@stbrody yeah that's fair. Just added you because it was relevant to me wanting to block merges into CAS develop branch. I'm going to merge this because it is just dead code at this point and there's no point in running a CI job that we aren't using.

PR #320, opened 2021-03-11
fix(cas): Fix underflow error in gas price math

stbrody:
ether's `BigNumber` package doesn't like decimals, so this changes the math from multiplying by .01 to dividing by 10

### DIFF
--- a/src/services/blockchain/__tests__/eth-bc-service.test.ts
+++ b/src/services/blockchain/__tests__/eth-bc-service.test.ts
@@ -82,9 +82,9 @@ describe('ETH service',  () => {
 
   test('gas price increase math', () => {
     const currentGas = BigNumber.from(1000)
-    expect(EthereumBlockchainService.increaseGasPrice(currentGas, 0)).toEqual(currentGas)
-    expect(EthereumBlockchainService.increaseGasPrice(currentGas, 1)).toEqual(BigNumber.from(1100))
-    expect(EthereumBlockchainService.increaseGasPrice(currentGas, 2)).toEqual(BigNumber.from(1200))
+    expect(EthereumBlockchainService.increaseGasPricePerAttempt(currentGas, 0)).toEqual(currentGas)
+    expect(EthereumBlockchainService.increaseGasPricePerAttempt(currentGas, 1)).toEqual(BigNumber.from(1100))
+    expect(EthereumBlockchainService.increaseGasPricePerAttempt(currentGas, 2)).toEqual(BigNumber.from(1200))
   })
 
 });
--- a/src/services/blockchain/__tests__/eth-bc-service.test.ts
+++ b/src/services/blockchain/__tests__/eth-bc-service.test.ts
@@ -10,6 +10,7 @@ import { container } from "tsyringe";
 
 import BlockchainService from "../blockchain-service";
 import EthereumBlockchainService from "../ethereum/ethereum-blockchain-service";
+import { BigNumber } from 'ethers';
 
 let ganacheServer: any = null;
 let ethBc: BlockchainService = null;
@@ -48,6 +49,12 @@ describe('ETH service',  () => {
     await done
   });
 
+  afterAll(async (done) => {
+    logger.imp(`Closing local Ethereum blockchain instance...`);
+    ganacheServer.close();
+    done();
+  });
+
   test('should connect to local ganache server', async () => {
     await ethBc.connect();
   });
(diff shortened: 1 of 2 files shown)


src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -76,14 +76,23 @@ export default class EthereumBlockchainService implements BlockchainService {
     } else {
       const gasPriceEstimate = await this.provider.getGasPrice();
```
valmack, line 77: Can we add a comment or something here for the units? I think this gives back the amount in "wei".

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -76,14 +76,23 @@ export default class EthereumBlockchainService implements BlockchainService {
...
       // Add 10% extra to gas price for each subsequent attempt
-      txData.gasPrice = gasPriceEstimate.add(gasPriceEstimate.mul(attempt * .01));
+      txData.gasPrice = EthereumBlockchainService.increaseGasPrice(gasPriceEstimate, attempt)
       logger.debug('Estimated Gas price: ' + txData.gasPrice.toString());
```
valmack, line 80: Can we add the denomination in this comment too? "wei"

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -76,14 +76,23 @@ export default class EthereumBlockchainService implements BlockchainService {
...
     }
+  }
 
-
+  /**
+   * Takes current gas price and attempt number and returns new gas price with 10% increase per attempt
+   * @param currentGas
+   * @param attempt
+   */
+  static increaseGasPrice(currentGas: BigNumber, attempt: number): BigNumber {
```
valmack, line 92: Can we change method name to something like `updateGasPriceByAttempt` ?

src/services/blockchain/__tests__/eth-bc-service.test.ts
```diff
@@ -73,10 +80,11 @@ describe('ETH service',  () => {
     expect(chainId).toEqual("eip155:1337")
   });
 
-  afterAll(async (done) => {
-    logger.imp(`Closing local Ethereum blockchain instance...`);
-    ganacheServer.close();
-    done();
-  });
+  test('gas price increase math', () => {
```
valmack, line 83: 👍

valmack (commented):
Small suggestions that would be helpful!

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stbrody: I'd actually rather not since it's not strictly relevant to our code and is really a property of the ethers library.  I don't feel like we should be trying too hard to document the behavior of libraries we depend on, or else we risk getting out of sync with them.  For example, what if over time deflation causes ethers to change what unit is returned from this function, and now our comments are out of date?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stbrody: Renamed to `increaseGasPricePerAttempt`

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
valmack: To do math properly we need to know the units though. And when we pass the value to a transaction we need to ensure it is wei not gwei, independent of what library we use.

It would be great if we had types for the denominations but I haven't seen a library that does that yet.

stbrody: > To do math properly we need to know the units though

How so?  10% is 10% regardless of units.

> when we pass the value to a transaction we need to ensure it is wei not gwei, independent of what library we use.

I would assume that if we're using the `ethers` library for everything, then as long as it's internally consistent we can pass around whatever values it gives us without needing to know or care what the underlying units are

valmack: Yeah I strongly dislike the assumption and I'm not positive it will be true over time. But I'm fine if you don't want to add a comment bc agree that could get stale in the future. Would push to include it in debug log though plz.

stbrody: Alright, I'm still worried about enshrining assumptions about upstream packages into our code, but I don't feel *that* strongly about it, so I'll go ahead and add the comment and log message for now

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stbrody: done

valmack (approved):
🙏

PR #327, opened 2021-03-15
hotfix: timeout on document load after 60s

oed:
Don't stall forever if the document doesn't load for whatever reason.

### DIFF
--- a/src/services/ceramic-service.ts
+++ b/src/services/ceramic-service.ts
@@ -27,7 +27,8 @@ export default class CeramicServiceImpl implements CeramicService {
   }
 
   async loadDocument<T extends Doctype>(docId: DocID): Promise<T> {
-    return await this._client.loadDocument(docId, {sync: false})
+    const docPromise = this._client.loadDocument(docId, {sync: false})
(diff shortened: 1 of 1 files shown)


src/services/ceramic-service.ts
```diff
@@ -27,7 +27,8 @@ export default class CeramicServiceImpl implements CeramicService {
   }
 
   async loadDocument<T extends Doctype>(docId: DocID): Promise<T> {
-    return await this._client.loadDocument(docId, {sync: false})
+    const docPromise = this._client.loadDocument(docId, {sync: false})
+    const timeoutPromise = new Promise((_, reject) => setTimeout(() => reject(`Timed out loading docid: ${docId.toString()}`), 60 * 1000))
```
stbrody, line 31: How does the failure manifest if the timeout fires and rejects the promise? Is an exception thrown?

valmack (approved):
LGTM!

src/services/ceramic-service.ts
oed: yes

PR #328, opened 2021-03-15
fix(ci): build docker image on main branch

valmack:
This is needed because CAS clay is still using the ECR image tagged develop. So this builds the ECR image for it to use when changes are merged into `main` so that the clay instances get the new image.

This code will be removed once we clean up the pipelines here.


oed (commented):
lgtm

PR #330, opened 2021-03-17
Hotfix: Exit anchor mode

valmack:
### Motivation

Clay CAS instances are hanging on our infra (AWS ECS) even though it seems like all the node processes should have exited. Testing locally if loading a doc fails execution would continue but the process would not exit until after the timeout of 60s. However, on our infra it seems that the introduction of this timeout keeps the process running indefinitely.

I believe this change helps the issue by clearing the timeout (because we aren't seeing this issue on dev), but I can't be sure until we run it on ECS. If it still hangs it is likely we need to make an explicit call to `process.exit()` after anchoring.

### Changes

- Clears `setTimeout` if `loadDocument` resolves or rejects
- Cleans up the execute anchor code to be a bit neater
  - *We are seeing out of order log messages because the anchor service module was being awaited when it should be run non-blocking*

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -167,7 +167,7 @@ export default class CeramicAnchorApp {
    * Execute a single anchor process
    */
   private async _executeAnchor(): Promise<void> {
-    await this.startWithConnectionHandling(async () => {
+    this.startWithConnectionHandling(async () => {
       const anchorService: AnchorService = container.resolve<AnchorService>('anchorService');
       await anchorService.anchorRequests();
     });
--- a/src/app.ts
+++ b/src/app.ts
@@ -141,7 +141,10 @@ export default class CeramicAnchorApp {
     const ipfsService: IpfsServiceImpl = container.resolve<IpfsServiceImpl>('ipfsService');
     await ipfsService.init();
 
-    await this._executeAnchor();
+    this.startWithConnectionHandling(async () => {
+      const anchorService: AnchorService = container.resolve<AnchorService>('anchorService');
+      await anchorService.anchorRequests();
+    });
   }
 
   /**
@@ -161,16 +164,6 @@ export default class CeramicAnchorApp {
       process.exit(1);
     }
     await fn();
-  }
-
-  /**
-   * Execute a single anchor process
-   */
-  private async _executeAnchor(): Promise<void> {
-    this.startWithConnectionHandling(async () => {
-      const anchorService: AnchorService = container.resolve<AnchorService>('anchorService');
-      await anchorService.anchorRequests();
-    });
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/app.ts
```diff
@@ -141,7 +141,10 @@ export default class CeramicAnchorApp {
     const ipfsService: IpfsServiceImpl = container.resolve<IpfsServiceImpl>('ipfsService');
     await ipfsService.init();
 
-    await this._executeAnchor();
+    this.startWithConnectionHandling(async () => {
```
stbrody, line 144: if we don't wait for the anchor to finish, won't we terminate the process before the anchor has run?

src/services/ceramic-service.ts
```diff
@@ -27,8 +27,24 @@ export default class CeramicServiceImpl implements CeramicService {
   }
 
   async loadDocument<T extends Doctype>(docId: DocID): Promise<T> {
+    let timeout: NodeJS.Timeout;
+
     const docPromise = this._client.loadDocument(docId, {sync: false})
-    const timeoutPromise = new Promise((_, reject) => setTimeout(() => reject(`Timed out loading docid: ${docId.toString()}`), 60 * 1000))
+      .then((data) => {
```
stbrody, line 33: I think you can use `.finally()` to avoid needing both a `then` and `catch` handler

src/app.ts
valmack: No bc the event loop will not be finished until the anchor is complete.

Code sample similar to what we do:
```js
async function stall() {
  await _start()
  console.log('stall')
}

async function _start() {
  console.log('_start')
  _nonBlocking()
}

async function _nonBlocking() {
  console.log('_nonBlocking')
  setTimeout(() => console.log('timed out'), 1000)
}

stall()
```

I.e. `await` is blocking relative to other code in the program, but the absence of it does not terminate the process

src/services/ceramic-service.ts
valmack: Cool ty!

src/services/ceramic-service.ts
```diff
@@ -27,8 +27,25 @@ export default class CeramicServiceImpl implements CeramicService {
...
+        return data;
+      })
+      .catch((err) => {
```
stbrody, line 36: Haha, now that you're using `finally` I don't think you need the `.then()` or `.catch()` cases at all

valmack: @stbrody Still do to return data and catch errors. Finally doesn't receive any args.

stbrody: test.js:
```
const foo = async function() {
    return "foo";
}

const runTest = async function() {
    const prom = foo()
    prom.finally(() => {
        console.log('finally called')
    });

    const res = await prom
    console.log("result: " + res)
}

runTest()
```

```
> node test.js
finally called
result: foo
```

stbrody: Finally just transforms the Promise into one that runs the callback before returning, but it doesn't change the way the resolved value or error passes through

valmack: Ahh I see. Updated

stbrody (approved):
lgtm

PR #335, opened 2021-03-17
fix(cas): Ensure gas price on retries is at least 10% more than previous attempt

stbrody:
builds on https://github.com/ceramicnetwork/ceramic-anchor-service/pull/334

### DIFF
--- a/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
+++ b/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
@@ -65,7 +65,7 @@ export default class EthereumBlockchainService implements BlockchainService {
    *   the gas price we set by a 10% multiple with each subsequent attempt
    * @private
    */
-  private async _setGasPrice(txData: TransactionRequest, attempt: number): Promise<void> {
+  async setGasPrice(txData: TransactionRequest, attempt: number): Promise<void> {
     const { overrideGasConfig } = config.blockchain.connectors.ethereum;
     if (config.blockchain.connectors.ethereum.overrideGasConfig) {
       txData.gasPrice = BigNumber.from(config.blockchain.connectors.ethereum.gasPrice);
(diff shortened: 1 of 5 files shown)


src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -90,14 +91,26 @@ export default class EthereumBlockchainService implements BlockchainService {
...
+   * previous attempt's gas, even if the gas price on chain has gone down since then. This is because
+   * retries of the same transaction (using the same nonce) need to have a gas price at least 10% higher
+   * than any previous attempts or the transaction will fail.
+   * @param currentGasEstimate
    * @param attempt
+   * @param previousGas
    */
-  static increaseGasPricePerAttempt(currentGas: BigNumber, attempt: number): BigNumber {
-    const tenPercent = currentGas.div(10)
+  static increaseGasPricePerAttempt(currentGasEstimate: BigNumber, attempt: number, previousGas: BigNumberish | undefined): BigNumber {
```
valmack, line 103: Why "BigNumberish" here? instead of BigNumber

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -90,14 +91,26 @@ export default class EthereumBlockchainService implements BlockchainService {
...
+    const tenPercent = currentGasEstimate.div(10)
     const additionalGas = tenPercent.mul(attempt)
-    return currentGas.add(additionalGas)
+    const newGas = currentGasEstimate.add(additionalGas)
+    if (attempt == 0 || !previousGas) {
```
valmack, line 107: Can `previousGas` here be 0 or 1? Is this better as `previousGas != undefined`?

valmack (approved):
LGTM. Pending parent PR approval

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stbrody: Because that's what `TransactionRequest.gasPrice` is :frowning_face: https://github.com/ethers-io/ethers.js/blob/6c43e20e7a68f3f5a141c74527ec63d9fe8458be/packages/abstract-provider/src.ts/index.ts#L24
I tried converting it at the call site, but then you have to put the handling for `undefined` there too which complicated things.  This seemed the most straightforward.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stbrody: I imagine gas can never be 0, and it's hard to imagine it ever being 1, but doesn't hurt to be precise.  Updated.

PR #338, opened 2021-03-19
Check running anchor process

valmack:
### Motivation

Having multiple CAS anchor modes running leads to failed batches.

### Changes

This introduces a new Dockerfile that allows us to run a script to check if AWS ECS has any anchor processes running. If so we exit without starting another CAS anchor process. Otherwise we proceed with starting CAS in anchor mode.

### Next steps

Build this image in CI and push it to AWS ECR

### DIFF
--- a/runner.sh
+++ b/runner.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 export TASKS="$(node ./runner/check-aws-ecs-tasks.js)"
 
@@ -6,8 +6,14 @@ if [[ $TASKS == 0 ]]; then
   cd $CAS_PATH && npm run start
   exit_code=$?
   if [[ $exit_code != 0 ]]; then
-    echo 'failed'
+    echo "Service exited with non-zero exit code"
   fi
 else
+  exit_code=$?
+  if [[ $exit_code != 0 ]]; then
+    echo "Failed to retrieve running tasks"
+  else
+  echo "Service is already running tasks"
   echo $TASKS
+  fi
 fi
--- a/runner.sh
+++ b/runner.sh
@@ -5,8 +5,8 @@ echo $TASKS
 
 if [[ $TASKS == "No running tasks found" ]]; then
   cd $CAS_PATH && npm run start
-  EC=$?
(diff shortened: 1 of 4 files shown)


runner.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+export TASKS="$(node ./runner/check-aws-ecs-tasks.js)"
+
+if [[ $TASKS == 0 ]]; then
```
stbrody, line 5: Sorry, my bash-fu isn't great.  What is this line checking?  At first I thought it was checking the exit code but then I see that is done below

runner.sh
```diff
@@ -1,19 +1,13 @@
...
 
 export TASKS="$(node ./runner/check-aws-ecs-tasks.js)"
+echo $TASKS
 
-if [[ $TASKS == 0 ]]; then
+if [[ $TASKS == "No running tasks found" ]]; then
   cd $CAS_PATH && npm run start
-  exit_code=$?
-  if [[ $exit_code != 0 ]]; then
+  EC=$?
```
stbrody, line 8: why the change from `exit_code` to `EC`?  I tend to prefer more meaningful names, even if they are a bit more verbose

runner/check-aws-ecs-tasks.js
```diff
@@ -17,23 +17,30 @@ async function main() {
...
     } else {
       if (data.taskArns.length > 0) {
+        console.log('Running tasks')
         console.log(data.taskArns)
       } else {
-        console.log(0)
+        console.log('No running tasks found')
       }
     }
   } catch (error) {
```
stbrody, line 33: Can any of this code within the try-catch actually throw an exception, other than the exception that is manually thrown on line 24?  Can we just log there instead and remove the try-catch?

stbrody: Also, this whole function is wrapped in a catch from how main() is run, so even if we kept throwing there the try-catch here still seems unnecessary

stbrody (approved):
A couple of small suggestions around code style/cleanup, but overall LGTM

PR #339, opened 2021-03-19
feat(ci): add workflow to build and push ecr image

valmack:
### Motivation

1. We are moving away from using CircleCI so this will replace the existing workflow that we have configured there
2. We want to build a new docker image for CAS that uses the "CAS runner" (see: #338)

### Changes

- Adds github workflow that builds image on push or when manually triggered

### Notes

- Have a feature branch in the workflow so we can test this without needing to merge to develop each time, but this needs to be merged in first on develop in order for the action to be available for testing on other branches. (Just how github actions works).

### DIFF
--- a/.github/workflows/build_and_push_ecr.yml
+++ b/.github/workflows/build_and_push_ecr.yml
@@ -2,6 +2,7 @@ on:
   push:
     branches:
       - main
+      - develop
       - feat/check-running-anchor-process # will remove after testing
   workflow_dispatch:
 
--- a/.github/workflows/build_and_push_ecr.yml
+++ b/.github/workflows/build_and_push_ecr.yml
@@ -1,0 +1,50 @@
+on:
+  push:
+    branches:
+      - main
+      - feat/check-running-anchor-process # will remove after testing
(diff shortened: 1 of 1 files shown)


stbrody:
> We are moving away from using CircleCI 

We are?  For tests too?  To what?  I find the way the integration tests surface their results much more annoying to work with than CircleCI.  Or are you just talking about for how the docker images are built?

stbrody (approved):

valmack:
> We are? For tests too? To what? I find the way the integration tests surface their results much more annoying to work with than CircleCI. Or are you just talking about for how the docker images are built?

Not tests for now but eventually we want to just use a single platform

stbrody:
> > We are? For tests too? To what? I find the way the integration tests surface their results much more annoying to work with than CircleCI. Or are you just talking about for how the docker images are built?
> 
> Not tests for now but eventually we want to just use a single platform

Do we know yet what the platform will be/what the UX will be like?  I much prefer a graphical interface like CircleCI, ideally integrated with the github UI but email notifications would work too.  I hate having it all through discord

PR #343, opened 2021-03-26
fix: run anchor mode if only 1 task is running

valmack:
When we detect if a task is running it is seeing itself running then exiting. So we need to check if any *additional* tasks are running.

### DIFF
--- a/runner/check-aws-ecs-tasks.js
+++ b/runner/check-aws-ecs-tasks.js
@@ -22,11 +22,11 @@ async function main() {
   if (data.$metadata.httpStatusCode > 399) {
     throw Error(data.$metadata.httpStatusCode)
   } else {
-    if (data.taskArns.length > 0) {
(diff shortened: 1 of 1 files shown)


runner/check-aws-ecs-tasks.js
```diff
@@ -22,11 +22,11 @@ async function main() {
...
     throw Error(data.$metadata.httpStatusCode)
   } else {
-    if (data.taskArns.length > 0) {
-      console.log('Tasks already running')
+    if (data.taskArns.length > 1) {
+      console.log('More than one task already running')
       console.log(data.taskArns)
     } else {
-      console.log('No running tasks found')
+      console.log('Only one running task found')
```
stbrody, line 29:
```suggestion
      console.log('Only one running task found (assumed to be self)')
```

stbrody (approved):
LGTM with one small logging suggestion

PR #350, opened 2021-03-29
Hotfix exit anchor mode

valmack:
### Motivation

CAS still hanging on clay when it attempts to retrieve a document and a timeout is exceeded.

### Changes
- Force exits after running the `anchorRequests` method in anchor mode
- Updates the docker-compose file so that anchor and server modes can be tested in containers

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,49 +1,102 @@
-# Docker Compose file Reference (https://docs.docker.com/compose/compose-file/)
-
 version: '3.4'
 
-# Define services
 services:
-#### Uncomment section bellow if you want to start the latest go-ipfs instance
-  cas:
-    # Configuration for building the docker image for the service
-    build:
-      context: . # Use an image built from the specified dockerfile in the current directory.
-      dockerfile: Dockerfile
-    restart: unless-stopped
-    environment: # Pass environment variables to the service
+
+  cas_api:
+    build: .
+    environment:
       - NODE_ENV=dev
       - APP_MODE=server
       - APP_PORT=8081
-      - ANCHOR_SCHEDULE_EXPRESSION=0 0 */1 * * *
-      - IPFS_API_URL=ipfs:5001
+
+      - ANCHOR_CONTROLLER_ENABLED=false
+      - ANCHOR_EXPIRATION_PERIOD=0
+      - ANCHOR_SCHEDULE_EXPRESSION=0/1 * * * ? *
+
       - BLOCKCHAIN_CONNECTOR=ethereum
-      - ETH_NETWORK=ganache
+      - CERAMIC_API_URL=http://host.docker.internal:7007
       - ETH_GAS_LIMIT=4712388
       - ETH_GAS_PRICE=100000000000
-      - ETH_RPC_HOST=http://ganache
-      - ETH_RPC_PORT=8545
-      - ETH_WALLET_PK=0x06dd0990d19001c57eeea6d32e8fdeee40d3945962caf18c18c3930baa5a6ec9
+      - ETH_NETWORK=ropsten
+      - ETH_OVERRIDE_GAS_CONFIG=false
+      - ETH_WALLET_PK=0x16dd0990d19001c50eeea6d32e8fdeef40d3945962caf18c18c3930baa5a6ec9
+
+      - IPFS_API_URL=ipfs:5001
+      - IPFS_API_TIMEOUT=10000
+
       - LOG_LEVEL=debug
       - LOG_TO_FILES=true
-      - TYPEORM_CONNECTION=sqlite
-      - TYPEORM_HOST=localhost
+      - LOG_PATH=/usr/local/var/log
+      - MERKLE_DEPTH_LIMIT=0
+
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -1,49 +1,102 @@
...
+    depends_on:
+      - ipfs
+    ports:
+      - "8081:8081"
+    networks:
+      - internal
+
+  cas_anchor:
+    build: .
+    restart: unless-stopped
```
stbrody, line 56: It seems like the API above should have `restart: unless-stopped`, but maybe `cas_anchor` shouldn't?

docker-compose.yml
```diff
@@ -1,49 +1,102 @@
...
       - "/usr/local/var/log:/usr/local/var/log"
     depends_on:
       - ipfs
-      - ganache
     networks:
       - internal
+
   ipfs:
-    image: ipfs/go-ipfs:latest
+    image: ceramicnetwork/js-ipfs-ceramic
```
stbrody, line 99: I guess you're holding off to make the switch to ipfs-daemon for a later patch?

docker-compose.yml
```diff
@@ -1,49 +1,102 @@
...
       - LOG_LEVEL=debug
       - LOG_TO_FILES=true
-      - TYPEORM_CONNECTION=sqlite
-      - TYPEORM_HOST=localhost
+      - LOG_PATH=/usr/local/var/log
+      - MERKLE_DEPTH_LIMIT=0
+
+      - TYPEORM_CONNECTION=postgres
+      - TYPEORM_DATABASE=anchor_db
+      - TYPEORM_HOST=database
```
stbrody, line 34: What does this `host` field represent?  Not a DNS URL I guess?

src/services/anchor-service.ts
```diff
@@ -180,6 +180,7 @@ export default class AnchorService {
       logger.debug(`Successfully anchored CID ${candidate.cid.toString()} for document ${candidate.document.id.toString()}`)
     }
     logger.imp(`Service successfully anchored ${anchors.length} CIDs.`);
+    process.exit()
```
stbrody, line 183: It feels like this should be happening at a higher layer.  There are other ways that `anchorRequests` can be called, including periodically on a schedule [when in `bundled` mode](https://github.com/ceramicnetwork/ceramic-anchor-service/blob/857a5bb93475a265b6dfaa12a8e7d440dac209e5/src/services/scheduler-service.ts#L34), or [according to an incoming request](https://github.com/ceramicnetwork/ceramic-anchor-service/blob/d224299195dbac599db000fdd80f0adbeb1523f8/src/controllers/anchor-controller.ts#L22) and this would cause both of those to terminate the process prematurely.

Really this feels like this should happen at the top level, [here](https://github.com/ceramicnetwork/ceramic-anchor-service/blob/6286ffd671ab5a7bf97cc592ec5a509b9032cc8f/src/app.ts#L58).  But maybe that's too late in the process that you're worried it won't stop the hang?  When you see the hanging processes, do you still see the `Ceramic Anchor Service started in anchor mode` [log message](https://github.com/ceramicnetwork/ceramic-anchor-service/blob/6286ffd671ab5a7bf97cc592ec5a509b9032cc8f/src/app.ts#L59)?

stbrody (commented):
A few questions

src/services/anchor-service.ts
valmack: I think the code you are pointing to is outdated but yeah you are right. Moving this to app.ts and will test in bundled mode too

docker-compose.yml
valmack: Docker has it's own internal networking so `database` in this case points to the database container

docker-compose.yml
valmack: Yup

src/services/anchor-service.ts
stbrody: Woah weird, not sure how I wound up on that version

docker-compose.yml
valmack: What this achieves is basically forcing cas to keep restarting after the process exits

For the api we don't need this feature because if the process exits it is not intentional

The wording of this option is a bit misleading unfortunately. It is more accurately "unless explicitly stopped by docker stop command"

stbrody: > For the api we don't need this feature because if the process exits it is not intentional

Right, that's why we'd want docker to restart, to make sure the service stays up in the event of a crash, right?

For the anchor mode, we don't want it to restart exactly *because* we expect it to terminate in normal operation, and it should only be started up again by the scheduler job on a specific schedule, no?

Maybe I'm misunderstanding how this docker file fits into the rest of our infra management story

valmack: No this docker-compose.yml is only used for testing locally. Not for testing infra.

stbrody: ah okay, then I guess it doesn't matter too much what the retry strategy here is.  Making the anchor process continually restart constantly is kind of weird behavior, but so is running it just once and never again, so I guess this is fine then

stbrody (approved):

PR #352, opened 2021-03-31
Releases

valmack:
### Motivation

Introduce releases here so it is easier to track changes going into Clay and Fire

### Changes
- Adds [release-it](https://github.com/release-it/release-it) package
- Cleaned up json spacing to 2 spaces instead of 3

### Notes

To make a release we just do `npm run release` and the CLI will provide prompts like the ones shown below.

This is a screenshot of what it will do (this was run doing `npm run release -- --dry-run` which we can always do for testing)

image


oed (approved):
Haven't used `release-it` before, but lgtm

stbrody (commented):
LGTM although it would be nice to update the README at some point with instructions on how the release process works

PR #376, opened 2021-04-28
chore: Bump ceramic deps

ukstv:


package.json
```diff
@@ -58,6 +55,7 @@
     "http-status-codes": "^2.1.4",
     "ipfs-core": "^0.3.0",
     "ipfs-http-client": "^48.1.2",
+    "key-did-resolver": "^1.1.2-rc.3",
```
oed, line 58: I assume this pkg is not needed since it's not imported anywhere?

oed (approved):
lgtm, would like to get a ✅ from @stbrody as well though.

package.json
ukstv: as a bunch of other packages

oed: What do you mean?

ukstv: base64url, config.ts, colors, ejs and few others are not used. Have a separate PR in the making to drop them.

ukstv: https://github.com/ceramicnetwork/ceramic-anchor-service/pull/377 @oed

stbrody (approved):
:rocket:

PR #379, opened 2021-04-29
feat(cas): Make CAS IPFS connect to pubsub

stbrody:


src/services/ipfs-service.ts
```diff
@@ -62,6 +62,10 @@ export class IpfsServiceImpl implements IpfsService {
       },
     });
 
+    // We have to subscribe to pubsub to keep ipfs connections alive.
+    // TODO Remove this when the underlying ipfs issue is fixed
+    await this._ipfs.pubsub.subscribe(config.ipfsConfig.pubsubTopic, () => { /* do nothing */ })
```
stbrody, line 67: Is this enough? In ceramic, we periodically re-subscribe every 15 seconds.  Do we need to do the same thing here?

@valmack @zachferland

oed (approved):

Issue #360, opened 2021-04-08
After nonce error check for success of all previous timed out txns, not just the most recent


stbrody:
See this log as an example: https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#logsV2:log-groups/log-group/$252Fecs$252Fceramic-dev-cas/log-events/cas_anchor$252Fcas_anchor$252F3dc940816ba54b8795418a992905af93

0xc364e6ec282d966c5092dd99075d7f0163d53565447c6ada73a9f60223f6eac5 was actually mined, but 0x54a6eee95ff4b76e75fc1593351461b77e1d4fcdd5cdacbab96ff4fddd6aa6eb was the most recent timed out attempt, and 0x63aad062438e48a2c7cea2d540b094cc5463f837386959ed7197805177e30f81 failed with a nonceExpired error

stbrody:
As part of this we may want to separate waiting for the transaction to be mined from waiting for some number of block confirmations

stbrody:
> As part of this we may want to separate waiting for the transaction to be mined from waiting for some number of block confirmations

I do think this could be done, but I decided not to do it as part of this ticket.  We can revisit this if we start to have issues with our anchor batches taking too long to complete.

PR #380, opened 2021-04-29
fix: Wait for more blocks before confirming anchor

oed:


oed:
We are adding a fix in CAS instead.

stbrody:
> We are adding a fix in CAS instead.

What does this mean?

We still occasionally see the blockTimestamp mismatch error, this was what we planned on doing to fix that.  Once we get the tests stable, it seems like we'll want to re-open this and merge it in

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -181,7 +182,7 @@ export default class EthereumBlockchainService implements BlockchainService {
     }
 
     const txReceipt: providers.TransactionReceipt = await this.wallet.provider.waitForTransaction(
-      txResponse.hash, 1, transactionTimeoutSecs * 1000);
+      txResponse.hash, NUM_BLOCKS_TO_WAIT, transactionTimeoutSecs * 1000);
```
stbrody, line 185: Will we need to increase the `transactionTimeoutSecs` proportionally?  Depends on how the ethers provider works and how it applies that timeout. Does the timeout just apply for waiting for it to be initially mined?  Or does it apply to the entire `waitForTransaction` operation including waiting for extra block confirmations?

oed: https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-waitForTransaction

Timeout is the frequency of queries. Confirmations is how many blocks need to be mined before the function returns.

https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/base-provider.ts#L927

stbrody: > Timeout is the frequency of queries. 

I don't really know what that means.  But if I'm reading the code from ethers right it looks like the timeout applies to the entire `waitForTransaction` call, including waiting for additional block confirmations.  That means we probably need to increase the timeout we're setting here proportional to the number of additional blocks we're waiting on.

stbrody: to clarify, this probably isn't a code change, but rather a change in the environment variable we set for `transactionTimeoutSecs`.  Which may in turn have ripple effects on how frequently we can anchor, and on the timeouts we set in the integration tests

oed: Oh i though `Timeout` specified poll interval.

stbrody: Naw, that's [provider.pollingInterval](https://github.com/ceramicnetwork/ceramic-anchor-service/blob/cceb3e5cbf863b472d9fae4c33bca18239fd6ebd/src/services/blockchain/ethereum/ethereum-blockchain-service.ts#L44)

stbrody (approved):
LGTM, I think probably the only change we should make now is changing the `ETH_TXN_TIMEOUT` env var from 180 (3 minutes) to 240 (4 minutes), adding 1 extra minute for 4 block confirmations (one every 15 seconds). @valmack 

I pulled https://github.com/ceramicnetwork/ceramic-anchor-service/issues/360 into next sprint and added an idea to clean up and separate waiting for mining from waiting for block confirmations, which I think will help improve the stability and reliability of the CAS.  But I think for now we can just do the small change to handle the block reorg issue and that'll probably be enough to get our tests green for now.

oed:
@stbrody do you want me to make that update in this PR?

stbrody:
> @stbrody do you want me to make that update in this PR?

I think it makes sense to get this in quickly and do #360 shortly thereafter

PR #394, opened 2021-05-10
Fix anchor hanging

valmack:


oed (approved):
Lgtm, I guess this should fix the errors from last night as well?

PR #406, opened 2021-05-14
chore(cas): Improve logging when loading a stream fails

stbrody:


src/services/ceramic-service.ts
```diff
@@ -33,7 +33,7 @@ export default class CeramicServiceImpl implements CeramicService {
 
     const timeoutPromise = new Promise((_, reject) => {
       timeout = setTimeout(() => {
-        reject(`Timed out loading docid: ${docId.toString()}`)
+        reject(new Error(`Timed out loading docid: ${docId.toString()}`))
```
stbrody, line 36: @ukstv, is this the right way to reject a promise so that the caller that is `await`ing the promise sees an `Error` thrown?

ukstv: yup

ukstv (approved):

PR #421, opened 2021-05-18
fix(cas): Estimate gas based on the txn request

stbrody:
It looks like we've been using the `ethers` library wrong in terms of how we've been doing our gas price estimation. We've been using the [getGasPrice](https://github.com/ethers-io/ethers.js/blob/d3b7130ed6ec50b192eb7f33905eaa327d65eee2/packages/abstract-provider/src.ts/index.ts#L217) method which returns the gas price of the previous block (I think), but there is also an [estimateGas](https://github.com/ethers-io/ethers.js/blob/d3b7130ed6ec50b192eb7f33905eaa327d65eee2/packages/abstract-provider/src.ts/index.ts#L228) method which takes information about the transaction you are trying to perform and explicitly estimates the amount of gas needed for it.


stbrody:
Nevermind, I think I misunderstood the distinction between gasPrice and gasLimit

PR #430, opened 2021-05-25
chore(cas): Warn if the tip isn't what we expect when loading stream at CommitID

stbrody:


src/services/anchor-service.ts
```diff
@@ -326,6 +326,10 @@ export default class AnchorService {
         throw new Error(`No valid ceramic document found with docId ${docId.toString()}`)
       }
 
+      if (doc.tip.toString() != request.cid) {
```
oed, line 329:
```suggestion
      if (doc.tip.toString() != request.cid.toString()) {
```
To be consistent with the logging below.

oed (commented):
Question.

stbrody:
> Question.

Question?

src/services/anchor-service.ts
stbrody: request.cid is already a string, not a CID object

src/services/anchor-service.ts
```diff
@@ -326,6 +326,10 @@ export default class AnchorService {
...
+        logger.warn(`When loading document ${doc.id.toString()} at commit ${request.cid.toString()}, doc was returned at tip ${doc.tip.toString()}`)
```
oed, line 330: Then why do we do toString here then?

oed (approved):

src/services/anchor-service.ts
stbrody: this is probably the part that should change I guess

stbrody: Okay, updated to remove the toString here

PR #432, opened 2021-05-26
fix(cas): Always confirm all previous txn attempts after nonce error

stbrody:
This most likely would have made last night's mainnet CAS anchor batch succeed


v-stickykeys (approved):
Ship!

PR #433, opened 2021-05-29
fix(cas): Don't use bad HTTP status when request doesn't exist

stbrody:
This makes Ceramic throw an exception rather than just return that the anchor failed, which complicates the error handling code paths

### DIFF
--- a/src/controllers/request-controller.ts
+++ b/src/controllers/request-controller.ts
@@ -40,7 +40,7 @@ export default class RequestController {
           const body = await this.#requestPresentation.body(request);
           return res.status(StatusCodes.OK).json(body);
         } else {
-          return res.status(StatusCodes.NOT_FOUND).send({
+          return res.status(StatusCodes.OK).send({
             error: "Request doesn't exist",
           });
         }


oed (approved):
As long as this doesn't break anything in js-ceramic currently.

stbrody:
My WIP commit for https://github.com/ceramicnetwork/js-ceramic/issues/1423 has code to change ceramic to interpret this error format correctly.  I'm going to hold off on merging this until that commit is merged first

oed:
@stbrody Should we wait for a week or more to make sure everyone running a Ceramic instance doesn't run into any issues?

stbrody:
> @stbrody Should we wait for a week or more to make sure everyone running a Ceramic instance doesn't run into any issues?

Okay, I just did some more testing and with this change, if this error is returned by the CAS, an *existing* version of js-ceramic gets an error like `Anchor failed for commit bagcqcerardaeqlrvdrsrn3uipmegqrhsbrafc6h5mii6wsbmrsyx73m3bvra of stream kjzl6cwe1jw146qwh9nslkhcj7cg7vc0q3b73zjw6sa5qnvqz0nxvcalw60gn1p: Unexpected status: undefined`

An updated version of Ceramic with my patch that is WIP [here](https://github.com/ceramicnetwork/js-ceramic/pull/1437) would get the more helpful error message `Anchor failed for commit bagcqceragmyjsgkwx25r6wrqzl5dufwoxmtv4zra4s76spu6k5k5swo7tmna of stream kjzl6cwe1jw14auzbbhsq75pj6i38myyvlfzyj4yo45zenh83g6zp7supjx20i1: Request doesn't exist`.

So having the newer version of the CAS with an older ceramic node means the error message is a bit less helpful, but the overall behavior is actually completely the same.

Given that, I'd like to go ahead and merge this in.

oed:
Thanks for making sure of this @stbrody  👍

ukstv:
So, on one hand Ceramic debates semantic HTTP status codes. On the other hand, we have this. Not good.

stbrody:
> So, on one hand Ceramic debates semantic HTTP status codes. On the other hand, we have this. Not good.

The problem is that right now Ceramic has no way to distinguish between errors *communicating* with the CAS and errors received *from* the CAS.  Errors *communicating* with the CAS (network errors, 503 Service Unavailable, etc) become thrown exceptions, but a successful response from the CAS that indicates that the CAS definitely knows that there's a problem with the request (anchor failed, request doesn't exist, etc) should go through a different code path.

Perhaps once we have a more robust error system in place where we can create categories of errors and do control flow based on those categories, then we could use semantic HTTP codes here.  In the meantime, this is by far the simplest way for Ceramic to distinguish what type of response it is receiving.

oed:
I think 404 errors are really bad to use. In browsers they are **always** outputted to console, regardless if you catch them or not. Makes it seem like things are broken.

v-stickykeys (commented):
To me it is two paradigms:

1 - You made a request that I can respond to but there are some problems with it so I return 200 and tell you the error (This is what you have written)

2 - You made a request that I can respond to but there are some problems with it so I return a machine readable error code and human readable error message

I very much prefer the latter because it makes development faster. As a dev I know what the issue is and can fix my code without asking whoever wrote the API to explain to me what I need to fix, i.e. good communication is minimal communication. This is probably a very RESTful perspective.

@oed in this case the 404 in the browser gives an accurate picture to machine and to users. Or is 404 inaccurate here?

ukstv:
Okay, I give up my objections here, as I see the overall benefit. I am still reluctant to consider 200 ok for errors here, but this can be addressed later.

oed:
> @oed in this case the 404 in the browser gives an accurate picture to machine and to users. Or is 404 inaccurate here?

It's not very accurate. These errors is an expected outcome in a lot of cases and it doesn't look good if there are errors in the console when the errors are expected.

oed:
One main problem for this is that people report these 404s as being bugs and think that they explain other issues they are seeing. This is really problematic if we get these 404 for errors we are expecting and it's part of normal operation.

ukstv:
It seems, that we are talking about different scenarios here. Fact: 404 on browser ends up dumped in console. But:
- nothing in the browser requests the anchor service,
- HTTP daemon returns 404 only in case of non-handled or invalid URL, a case of non-existent stream is a timeout, not 404.

Unless I miss something, I doubt the argument of 404 in browser console can be applied here.

oed:
> nothing in the browser requests the anchor service

If you run `@ceramicnetwork/core` in the browser it does!

v-stickykeys:
> One main problem for this is that people report these 404s as being bugs and think that they explain other issues they are seeing. This is really problematic if we get these 404 for errors we are expecting and it's part of normal operation.

This is fair enough. https://restfulapi.net/http-status-204-no-content/ please?

v-stickykeys (approved):
No reason to hold this up really

stbrody:
Going ahead with merging this.  We can always revisit this after the error handling project

PR #457, opened 2021-06-15
chore(cas): Renames to use stream terminology throughout code

stbrody:
One thing that still needs to be decided is what we're going to do about the [docid field](https://github.com/ceramicnetwork/ceramic-anchor-service/blob/886fd416f2b0c8d9f3bf011a58fc894d35be297a/src/merkle/merkle-objects.ts#L76) in the bloom filter in the merkle tree metadata. If we don't change it now, then we're stuck using `docid` literally forever, confusing the story when working with indexer partners and adding complication and confusion for the rest of time.  If we *do* change it now, then any streams created in the first week of ELP won't get indexed - at least until they are updated again and get a new AnchorCommit.  Personally I kind of think we should just make the cut now and tell our ELP cohort 1 to do a trivial update to any streams they want to get indexed.  The likelihood of those streams not getting indexed actually ever being an issue seems really low, not worth adding complexity to the system for the rest of all time over.

We could even make the change now and instruct indexing partners in the future to check for streamId first and fall back to checking for docId

@oed @michaelsena


stbrody:
FYI @v-stickykeys

src/controllers/request-controller.ts
```diff
@@ -63,17 +63,20 @@ export default class RequestController {
     try {
       logger.debug(`Create request ${JSON.stringify(req.body)}`);
 
-      const { cid, docId } = req.body;
+      const cid = req.body.cid;
+      // TODO docId check for backwards compat with old Ceramic nodes. Remove once all infra nodes
+      // updated to new Ceramic version
```
oed, line 68: We should probably keep this around for longer. Ppl are using their own nodes on clay. Not only tied to our infra.
```suggestion
      // TODO docId check for backwards compat with old Ceramic nodes. Remove once confident that
      // most nodes in the network updated to new Ceramic version
```

oed: Please also indicate which version here for future reference.

src/merkle/merkle-objects.ts
```diff
@@ -72,20 +72,20 @@ export class BloomMetadata implements MetadataFunction<Candidate, TreeMetadata>
   generateMetadata(leaves: Array<Node<Candidate>>): TreeMetadata {
     const bloomFilterEntries = new Set<string>()
     for (const node of leaves) {
-      const doc = node.data.document
-      bloomFilterEntries.add(`docid-${doc.id.baseID.toString()}`)
-      if (doc.metadata.schema) {
-        bloomFilterEntries.add(`schema-${doc.metadata.schema.toString()}`)
+      const stream = node.data.stream
+      bloomFilterEntries.add(`docid-${stream.id.baseID.toString()}`)
```
oed, line 76:
```suggestion
      bloomFilterEntries.add(`streamid-${stream.id.baseID.toString()}`)
```

oed (changes requested):

src/controllers/request-controller.ts
stbrody: I won't know the version number to use before merging this.  I need to get this merged, released, and deployed on our infra, then I can make the change to ceramic to use the new format and get that released.

I'll include a reference to the ceramic github issue that will handle the support, that seems like the best we can do

oed: I guess you could just include the js-ceramic version we are currently at? Or better, a reference to the issue for fixing it!

oed (approved):

src/controllers/request-controller.ts
oed: Ah, that's what you did 👍

PR #461, opened 2021-06-22
chore(cas): Do not rely on process wide globals

stbrody:
Instead of relying on the global dependency injection `container` object, pass the container explicitly.
Instead of relying on the global `config` object, register the config with the dependency injection container and resolve it wherever needed.

Not the cleanest, but seemed the shortest path to allowing multiple CAS instances to coexist in the same process space

### DIFF
--- a/src/repositories/request-repository.ts
+++ b/src/repositories/request-repository.ts
@@ -5,12 +5,17 @@ import { BaseRepository } from 'typeorm-transactional-cls-hooked';
 import { Request, RequestUpdateFields } from "../models/request";
 import { RequestStatus } from "../models/request-status";
 import { logEvent } from '../logger';
-import { config } from "node-config-ts";
-import { singleton } from "tsyringe";
+import { Config } from 'node-config-ts';
+import { inject, singleton } from 'tsyringe';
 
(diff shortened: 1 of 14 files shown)


src/repositories/request-repository.ts
```diff
@@ -5,13 +5,18 @@ import { BaseRepository } from 'typeorm-transactional-cls-hooked';
...
-import { singleton } from "tsyringe";
+import { Config } from 'node-config-ts';
+import { inject, singleton } from 'tsyringe';
 
 @singleton()
 @EntityRepository(Request)
 export default class RequestRepository extends BaseRepository<Request> {
 
+  constructor(
+    @inject('config') private config?: Config) {
```
ukstv, line 16: I am curious, do you know @stbrody why the argument has to be marked as optional here?

ukstv (approved):
🏇

src/repositories/request-repository.ts
stbrody: I was just copying what I saw done elsewhere.  I suspect it's for unit tests that might not choose to register all possible components?

PR #463, opened 2021-06-23
chore(cas): Pass database connection in to CeramicAnchorApp

stbrody:
I'm *pretty* sure I did this right, and the tests all pass, but this is my first time using TypeORM and I mostly just figured this out from reading the docs, so it would be good if there was someone with more familiarity with it to double check me.  I'm not actually sure who on the team has experience working with TypeORM, so please chime in if you do.  @ukstv @PaulLeCam @v-stickykeys @zachferland @oed @smrz2001

### DIFF
--- a/src/services/__tests__/db-connection.ts
+++ b/src/services/__tests__/db-connection.ts
@@ -1,4 +1,4 @@
-import { ConnectionOptions, createConnection, getConnection } from "typeorm";
+import { Connection, ConnectionOptions, createConnection, getConnection } from 'typeorm';
 import { Anchor } from "../../models/anchor";
 import { Request } from "../../models/request";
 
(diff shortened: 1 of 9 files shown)


src/services/__tests__/db-connection.ts
```diff
@@ -12,16 +12,15 @@ const sqliteConf : ConnectionOptions = {
 };
 
 const DBConnection = {
-  async create(): Promise<void> {
-    await createConnection(sqliteConf);
+  async create(): Promise<Connection> {
+    return await createConnection(sqliteConf);
```
ukstv, line 16: unnecessary await

ukstv (approved):

v-stickykeys (approved):
LGTM!

Would be nice to have tests for the transaction rollback to ensure that piece works as expected but given that it is likely a fairly standard construction seems ok.

PR #464, opened 2021-06-24
chore(cas): Clean up IPFS initialization

stbrody:
This meant that the cas only instructs its ipfs node to subscribe to pubsub when performing an anchor.

Ideally we should probably be making the CAS IPFS node subscribe to pubsub on its own at startup via an environment variable, rather than waiting for a CAS node to instruct it to subscribe. But this will at least add one more case where the ipfs node will get subscribed

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -104,6 +104,10 @@ export default class CeramicAnchorApp {
     return configCopy
   }
 
+  private _shouldStartIpfs(): Boolean {
+    return this.config.mode == "anchor" || this.config.mode == "bundled" || this.config.anchorControllerEnabled;
+  }
+
   /**
    * Start application
    */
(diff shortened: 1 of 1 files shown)


v-stickykeys (commented):
Why does it need to start up IPFS in server mode?

stbrody:
> Why does it need to start up IPFS in server mode?

It doesn't strictly *need* to, but it's harmless to, simplifies the code paths, and ensures that we subscribe the node to pubsub in more cases.  I'm thinking about a time where we restart/deploy to an environment but then don't do an anchor for 12 hours, that can leave the new IPFS node not connected to pubsub that whole time which seems like it could cause it to get disconnected from peers.

v-stickykeys:
@stbrody As of right now this is harmless, but in general we shouldn't have multiple ipfs clients pointed to the same ipfs daemon when using pubsub, because we can end up in states where one the client handles think they are subscribed when they are not.

For that reason I'd rather avoid this path, given that on our infra we run both modes as separate services but only one ipfs node for CAS.

If this change is worth it then it'd be great to just add a comment warning ourselves that pubsub subscription logic is unreliable in this state.

stbrody:
> For that reason I'd rather avoid this path, given that on our infra we run both modes as separate services but only one ipfs node for CAS.

Okay, I changed this so we once again only start up IPFS if it's actually needed, and renamed the PR to clarify that this is mostly just code cleanup now.

v-stickykeys (approved):
Ty 👍

PR #468, opened 2021-06-30
chore(cas): Test CAS consensus behavior

stbrody:


stbrody:
This test is now passing consistently in my IDE and in the terminal, but is failing with what looks like some kind of babel issue in CircleCI for some reason.  Any ideas @ukstv @PaulLeCam?

```
 FAIL  src/__tests__/ceramic_integration.test.ts
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/en/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /home/circleci/cas/node_modules/ipfs-http-server/src/index.js:140
        const enableCors = Boolean(cors.origin?.length)
```

PaulLeCam:
@stbrody is CI using a different version of Node than the one you're using locally?
The issue seems to come from the use of `?.` that was added in ES2019 IIRC, so older versions of Node (before v14 maybe?) might not support it.

stbrody:
> @stbrody is CI using a different version of Node than the one you're using locally?
> The issue seems to come from the use of `?.` that was added in ES2019 IIRC, so older versions of Node (before v14 maybe?) might not support it.

Yeah that would make sense, thanks @PaulLeCam.  Any idea how to tell what version of node CircleCI is using @v-stickykeys?  I do see in the "Spin up environment" step of the circleci logs:
```
Starting container circleci/node:12
circleci/node:12:
  using image circleci/node@sha256:29503845858963005b42d2d34e972cb08abab18bb67cbb265a579432fe4144c5
  pull stats: Image was already available so the image was not pulled
  time to create container: 72ms
Warning: No authentication provided, using CircleCI credentials for pulls from Docker Hub.
  image is cached as circleci/node:12, but refreshing...
12: Pulling from circleci/node
```

Does the fact that the container is called `circleci/node:12` imply that it's using node version 12?  If so, can we get that updated to version 14?

PaulLeCam:
> > @stbrody is CI using a different version of Node than the one you're using locally?
> > The issue seems to come from the use of `?.` that was added in ES2019 IIRC, so older versions of Node (before v14 maybe?) might not support it.
> 
> Yeah that would make sense, thanks @PaulLeCam. Any idea how to tell what version of node CircleCI is using @v-stickykeys? I do see in the "Spin up environment" step of the circleci logs:
> 
> ```
> Starting container circleci/node:12
> circleci/node:12:
>   using image circleci/node@sha256:29503845858963005b42d2d34e972cb08abab18bb67cbb265a579432fe4144c5
>   pull stats: Image was already available so the image was not pulled
>   time to create container: 72ms
> Warning: No authentication provided, using CircleCI credentials for pulls from Docker Hub.
>   image is cached as circleci/node:12, but refreshing...
> 12: Pulling from circleci/node
> ```
> 
> Does the fact that the container is called `circleci/node:12` imply that it's using node version 12? If so, can we get that updated to version 14?

Yes AFAIK it would use the latest version of node 12, if you replace by `circleci/node:14` it should update.

src/__tests__/ceramic_integration.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    // At this point we have verified that *all* involved ceramic nodes (both user-facing ceramic
+    // nodes as well as both ceramic nodes used by the 2 anchor services) all know about the newest
+    // version of the stream.
+    await anchorUpdate(doc1, cas1)
+    expect(doc1.state.anchorStatus).toEqual(AnchorStatus.ANCHORED)
+
+    // This fails! Even though all Ceramic nodes know about the new update, 'cas1' still anchors
+    // the genesis contents that it was told to anchor.
+    // TODO(#253): Uncomment this once CAS consensus rules are improved
+    //expect(doc1.content).toEqual(updatedContent)
```
stbrody, line 327: this is the actual thing I set out to reproduce with this test.  This is the piece that gets fixed with the CAS Consensus Improvements project.  The PR that actually fixes the behavior will uncomment this line

stbrody:
This is now ready for review @ukstv @v-stickykeys 

CC @oed @smrz2001

src/__tests__/ceramic_integration.test.ts
```diff
@@ -0,0 +1,330 @@
...
+  await Utils.delay(500)
+
+  const tillAnchored = stream
+    .pipe(
+      filter((state) => [AnchorStatus.ANCHORED, AnchorStatus.FAILED].includes(state.anchorStatus)),
+      take(1),
+    )
+    .toPromise();
+  await anchorService.anchor();
+  await tillAnchored;
```
v-stickykeys, line 133: jw is there a reason `tillAnchored` should be constructed before the call to `anchorService.anchor` is made?

src/__tests__/ceramic_integration.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    expect(doc1.state.anchorStatus).toEqual(AnchorStatus.ANCHORED)
+
+    // Now that genesis commit has been anchored do an update and make sure anchoring works again
+    await doc1.update({foo:2}, null, { anchor: true })
+    await anchorUpdate(doc1, cas1)
+    expect(doc1.state.anchorStatus).toEqual(AnchorStatus.ANCHORED)
+    expect(doc1.content).toEqual({foo:2})
+  }, 60 * 1000 * 3);
+
+  test('Consensus for anchors', async () => {
```
v-stickykeys, line 299: Can we rename this from `Consensus for anchors` to `Anchors latest update at time anchor is requested`?

src/__tests__/ceramic_integration.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+  beforeEach(async () => {
+    console.log(`Starting test: ${expect.getState().currentTestName}`);
+  })
+
+  afterEach(async () => {
+    console.log(`Finished test: ${expect.getState().currentTestName}`);
+  })
+
+  test('Anchors are independent', async () => {
```
v-stickykeys, line 267:
```suggestion
  test('Anchors on one CAS are not repeated on another', async () => {
```

src/__tests__/ceramic_integration.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    expect(doc2.state.anchorStatus).toEqual(AnchorStatus.PENDING)
+
+    // Now test that anchoring on CAS2 doesn't anchor requests made against CAS1
+    await doc1.update({foo:2}, null, { anchor: true })
+    await anchorUpdate(doc2, cas2)
+    expect(doc1.state.anchorStatus).toEqual(AnchorStatus.PENDING)
+    expect(doc2.state.anchorStatus).toEqual(AnchorStatus.ANCHORED)
+  }, 60 * 1000 * 3);
+
+  test('Multiple anchors for same stream', async () => {
```
v-stickykeys, line 286:
```suggestion
  test('Genesis commit and updates get anchored for same stream', async () => {
```

v-stickykeys (commented):
Would be helpful if it is more clear what each test is meant to do so that we can ensure we are actually achieving that and it will be easier to know how thorough/exhaustive the tests are.

src/__tests__/ceramic_integration.test.ts
stbrody: yes, have to have the listener registered before the anchor or else the anchor might complete before you start listening and then you wouldn't get any updates

stbrody: Oh wait, actually this isn't quite true anymore. Now that streams get fixed to the version they are loaded at, then even if the anchor had completed before you subscribed to changed you'd still get notified.  I can switch things around now that that's the case so the control flow is simpler

src/__tests__/ceramic_integration.test.ts
stbrody: That wording feels overly specific to me.  The real thing we are testing here is multiple anchors, the fact that one is a genesis commit and one is an update is just a result of what's easiest to test, but isn't the main point.  So I think I'd rather leave the description as-is

stbrody (commented):
> Would be helpful if it is more clear what each test is meant to do so that we can ensure we are actually achieving that and it will be easier to know how thorough/exhaustive the tests are.

Yeah sorry, I should have been more clear.  The first 2 tests are really just testing the test setup itself - making sure that running multiple CAS instances in-memory simultaneously is working. The last test is the one that's the actual test I set out to write to repro the issue described in https://github.com/3box/3box/issues/1300.
I added new "describe" sub-sections to the test to try to make this more clear. PTAL

v-stickykeys (approved):
LG ty!!

PR #473, opened 2021-07-08
feat(cas): Update CAS to always anchor most recent version of a Stream, as loaded from Ceramic

stbrody:


stbrody:
This one's a bit of a beast, so setting 3 reviewers to make sure it makes sense to multiple people and not just to me.

Also CC @smrz2001

src/services/anchor-service.ts
```diff
@@ -99,17 +104,28 @@ export default class AnchorService {
...
 
+    // filter valid requests
+    const numInitialRequests = requests.length
+    requests = requests.filter((req) => {
+      return candidates.find((candidate) => {
+        return candidate.acceptedRequests.find((req2) => {
+          return req2.id == req.id
+        })
+      })
+    })
```
oed, line 127: Is this potentially `O(n^3)`?

src/services/anchor-service.ts
```diff
@@ -240,188 +256,180 @@ export default class AnchorService {
...
-   * Selects which Candidate CID should be anchored for each streamId
-   * @param candidates - List of Candidates each representing one anchor request
-   * @return a tuple whose first element is an array of the Candidates that were selected for anchoring,
-   *   and whose second element is an array of Candidates that were rejected by the conflict resolution rules
+   * First starts by loading the CommitID for each pending Request on this Stream. This ensures
+   * that the Ceramic node we are using has at least heard of and considered every commit that
+   * has a pending anchor request. Then we load the current version of the Stream from that Ceramic
+   * node and can then trust that the version of the Stream we receive back is the best, most current
+   * version of the stream based on the node's internal conflict resolution mechanism. We can then
+   * use the current version of the Stream to decide what CID to anchor.
```
oed, line 382: Can we make this into a multiquery that loads both the *streamid* and the *commitid* in one request? We do this in 3id-did-resolver for example.

oed (changes requested):
This is a big change that is a little hard to review, mostly because I lack some context on the CAS code base.

I think we can improve it slightly by making a multiquery, see below.

src/services/anchor-service.ts
stbrody: We discussed this on the original spec document as well.  We can't load the CommitIDs and the StreamID in one multiquery as Ceramic processes all requests in a Multiquery concurrently, so there's no guarantee that the load of the base StreamID will consider all the tips from the CommitIDs.  We could, however, group all the CommitIDs into a single Multiquery and just have the base StreamID load happen after.  I can make that change

oed: > so there's no guarantee that the load of the base StreamID will consider all the tips from the CommitIDs

What? That seems like a bug?

stbrody: > What? That seems like a bug?

That's debatable.  It does make some sense for a Multiquery to load all of the component requests concurrently.  We could, however, change MultiQuery to always load all CommitIDs provided first, and then if there are also StreamIDs for the same streams only load them *after* the commitIDs for the same streams have been considered, but that could slow things down for use cases that don't require this behavior of using the CommitID load as a way to inform the node of a tip it may not have heard of.

oed: >  We could, however, change MultiQuery to always load all CommitIDs provided first

Why would it need to load CommitIDs first? If you load a CommitID after should we not do conflict resolution if the commit tip is not in the log? 

> but that could slow things down for use cases that don't require this behavior

Could we not have the multiquery system hold handles to the streams that they load and listen for updates until all streams are loaded?

stbrody: > Could we not have the multiquery system hold handles to the streams that they load and listen for updates until all streams are loaded?

Yeah actually something like that could work.  Syncing all loaded streams to the most recent state in the cache right before returning them.  I'll give that a shot on Monday

oed: Nice.

stbrody: So thinking about this a bit more, there's a cost to using MultiQueries right now, which I think exposes a bigger issue with the multiQueries api in general.

With the implementation as it's written right now in this PR, it's possible for there to be multiple requests for different tips on the same stream, and for some of those requests to fail while others succeed, and the best tip for the stream still gets anchored.  Now if I switch to using multiqueries, now if loading *any* of the CommmitIDs fails, the whole multiQuery fails, and the only option left is to fail *all* the requests on this stream.

This is an issue for the multiQuery api in general, I'll file a ticket for it.

In the meantime, I think we should merge this as-is.  The CAS isn't very latency sensitive, so a few extra round trips won't be killer, plus we're doing the processing for multiple streams concurrently anyway.

stbrody: oh actually nevermind - I guess multiQuery just swallows any error for any of the individual queries.  That removes this concern then.  I'll go ahead and try to switch over to multiqueries, and I'll file a ticket for more fine-grained error reporting from multiQueries

stbrody:
Okay, PR is updated, review comments addressed, and tests passing!  Reviewers please take another look

src/services/anchor-service.ts
stbrody: Updated!

src/services/anchor-service.ts
stbrody: Updated

oed (approved):
lgtm!

oed:
Would be good if someone else also looked over it, @ukstv @smrz2001

src/services/anchor-service.ts
```diff
@@ -123,22 +135,22 @@ export default class AnchorService {
...
     logger.debug("Persisting results to local database")
-    await this._persistAnchorResult(anchors)
+    await this._persistAnchorResult(anchors, acceptedRequests)
 
     logEvent.anchor({
       type: 'anchorRequests',
       requestIds: requests.map(r => r.id),
-      clashingRequestsCount: numInitialRequests - requests.length,
-      validRequestsCount: requests.length,
+      clashingRequestsCount: requests.length - acceptedRequests.length, // This field is poorly named
```
stbrody, line 147: Note @v-stickykeys, this pr changes the meaning of this `clashingRequestsCount` field so it's now kind of poorly named.  Previously it only reported requests rejected by conflict resolution, but now it will return a count of requests that failed for *any* reason including conflict resolution, networking errors, timeouts, etc.

I can rename the field but was afraid that would break the metrics dashboards - let me know if you think I should go ahead and rename it to `failedRequestsCount`

src/services/ceramic-service.ts
```diff
@@ -40,4 +41,21 @@ export default class CeramicServiceImpl implements CeramicService {
...
+
+    const queryPromise = this._client.multiQuery(queries)
+      .finally(() => {
+        clearTimeout(timeout);
+      });
+
+    const timeoutPromise = new Promise<Record<string, Stream>>((_, reject) => {
+      timeout = setTimeout(() => {
+        reject(new Error(`Timed out loading multiquery`))
+      }, 60 * 1000);
```
smrz2001, line 56: Is 60 seconds too much?

src/merkle/merkle-tree.ts
```diff
@@ -1,15 +1,17 @@
 import { CompareFunction, MergeFunction, MetadataFunction, Node, PathDirection, TreeMetadata } from './merkle';
 
 /**
- * Merkle tree structure
+ * Merkle tree structure.
+ * Type 'N' is the type of the nodes in the tree. Type 'L' is the type of the leaf nodes specifically,
```
smrz2001, line 5: Pardon my ignorance here - why was this necessary?

src/merkle/merkle-objects.ts
```diff
@@ -1,49 +1,187 @@
...
+    if (stream.state.anchorStatus == AnchorStatus.ANCHORED) {
+      this._alreadyAnchored = true
+    } else {
+      this._cid = stream.tip
+      this._metadata = stream.metadata
+    }
+
+    // Check the log of the Stream that was loaded from Ceramic to see which of the pending requests
+    // are for CIDs that are included in the current version of the Stream's log.
+    const includedRequests = this._requests.filter((req) => {
```
smrz2001, line 146: Might be able to reduce the number of loops and nesting here. Maybe not necessary right now but performance will continue to degrade per stream processed as the size of the stream log grows.

smrz2001: I mean the `filter`/`includes`/`find` calls between line 146 and line 166.

src/services/anchor-service.ts
v-stickykeys: We aren't tracking this in metrics anyway right now but the best thing to do is separate them and name them accordingly. So that would be 
```
conflictingRequestsCount
totalFailedRequestsCount
acceptedRequestsCount
```

etc..

src/merkle/merkle-objects.ts
stbrody: I think a stream log would be considered "large" if it had around a thousand entries in it, which is still going to be nothing for these loops.  So I'm not really worried about the perf cost here.  Especially since anchoring happens twice a day currently, this code is far from performance sensitive.  I'd rather avoid premature optimization until/unless we see an issue here

src/merkle/merkle-tree.ts
stbrody: As discussed on our call this morning, this is to allow more precise type usage and distinguishing between the type used for the leaf nodes from that used for the intermediate nodes in the tree

src/services/ceramic-service.ts
stbrody: I don't think so.  And with streams being loaded in concurrent batches of 128 streams at a time, one stream taking a long time to time out doesn't actually tie up the rest of the system very much.

src/services/anchor-service.ts
stbrody: sweet thanks, updated

smrz2001 (approved):
:shipit:

PR #474, opened 2021-07-29
Feat/build push restart

mahendra3web:
Test if workflow_dispatch input of manual environment selection for GitHub Actions works better in develop branch than feature branch  feat/build-push-restart


v-stickykeys (changes requested):
Is this merge just for testing? If so we should comment out the trigger on push to develop branch until we are sure this will work correctly

v-stickykeys (approved):
👍

PR #477, opened 2021-08-03
fix: Unmasking release-candidate and main push on branches

mahendra3web:
Tested both push on develop branch works for restart_cas and also with workflow_dispatch with dev environment selection.

### DIFF
--- a/.github/workflows/build_push_and_restart_cas.yml
+++ b/.github/workflows/build_push_and_restart_cas.yml
@@ -3,8 +3,8 @@ name: build_push_and_restart_cas
 on:
   push:
     branches:
-      ##MASKED## - main
-      ##MASKED## - "release-candidate"
+      - main
+      - "release-candidate"
       - develop
   workflow_dispatch:
     inputs:


.github/workflows/build_push_and_restart_cas.yml
```diff
@@ -3,8 +3,8 @@ name: build_push_and_restart_cas
 on:
   push:
     branches:
-      ##MASKED## - main
-      ##MASKED## - "release-candidate"
+      - main
+      - "release-candidate"
```
smrz2001, line 7: Are the double quotes needed? Not a huge deal, just looks a little odd since they're not used for the other branch names.

smrz2001: Also, at [L31](https://github.com/ceramicnetwork/ceramic-anchor-service/pull/477/files#diff-2a4bf41db23fc4c3523c50fbe7a38582e5eef6e930a028845cbf3b4aa52d7944R31) and [L35](https://github.com/ceramicnetwork/ceramic-anchor-service/pull/477/files#diff-2a4bf41db23fc4c3523c50fbe7a38582e5eef6e930a028845cbf3b4aa52d7944R35), are both `ref` and `base_ref` needed?

smrz2001: For [reference](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) regarding the `github` context.

v-stickykeys:
This looks good for now. I think we can hold on merging until adding the lock

PR #480, opened 2021-08-05
chore(cas): Don't anchor early when merkle tree full

stbrody:
Turns out this only happens if you are running in "bundled" mode so wasn't really affecting us much anyway, but taking it out just to be safe


stbrody:
ping @smrz2001

smrz2001 (approved):
🚀

PR #496, opened 2021-09-07
feat: Use 1559 blockchain transactions

ukstv:
Drop gasFee in favour of maxPriority and (when needed) maxFeePerGas.


src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -269,91 +268,93 @@ export default class EthereumBlockchainService implements BlockchainService {
...
-        return await this._confirmTransactionSuccess(txResponse, network, transactionTimeoutSecs)
-      } catch (err) {
-        logger.err(err)
-
-        const { code } = err
-        if (code) {
-          if (code === ErrorCode.INSUFFICIENT_FUNDS) {
-            const txCost = (txData.gasLimit as BigNumber).mul(txData.gasPrice)
-            if (txCost.gt(walletBalance)) {
+    return this.withWalletBalance(async (walletBalance) => {
```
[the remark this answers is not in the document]
stbrody: ugh I see now.  You're trying to clean up how we log the final balance when the function ends. I really hate the extra scope, and the use of passing in a function to call that isn't actually a callback and will be executed inline is weird to me as well.  Feels like a whole lot of unnecessary machinery.  But I do agree that the random lines of code after the while loop that have to be remembered, and that prevents an early-return anywhere in main body of the function, isn't ideal either.  Sigh this would be so much cleaner in so many other languages.  RAII with C++ or Rust, or a `defer` statement in Go, would all be so much nicer for this.

Ultimately, while I personally would probably have left the structure closer to what it was, it's not really clear cut which approach is cleaner, so this seems fine to me if you like it better this way.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -89,14 +95,24 @@ export default class EthereumBlockchainService implements BlockchainService {
...
-        gasPriceEstimate,
+      const prevMaxPriorityFeePerGas = BigNumber.from(txData.maxPriorityFeePerGas || 0)
+      const nextMaxPriorityFeePerGas = EthereumBlockchainService.increaseGasPricePerAttempt(
+        feeData,
         attempt,
-        txData.gasPrice
+        prevMaxPriorityFeePerGas
+      )
+      const difference = nextMaxPriorityFeePerGas.sub(prevMaxPriorityFeePerGas)
+      if (feeData.maxPriorityFeePerGas) {
```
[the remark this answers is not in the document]
stbrody: saw your comment below.  Is this also about working Ganache not supporting EIP1559?  I hate having production code that exists just to work around testing, if it can be avoided (although admittedly sometimes it can't be avoided).  Would it be possible to introduce something in the test setup that wraps Ganache and coerces it to an EIP1559-compatible API, so that the workaround lives entirely in test code not in production code?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -109,35 +125,31 @@ export default class EthereumBlockchainService implements BlockchainService {
...
   static increaseGasPricePerAttempt(
-    currentGasEstimate: BigNumber,
+    feeData: providers.FeeData,
     attempt: number,
     previousGas: BigNumberish | undefined
   ): BigNumber {
-    const tenPercent = currentGasEstimate.div(10)
-    const additionalGas = tenPercent.mul(attempt)
-    const newGas = currentGasEstimate.add(additionalGas)
+    // Ganache still does not support 1559, so here is a workaround
```
stbrody, line 137: see above comment.  Would prefer if we could handle this in test code instead of prod code if possible

stbrody (commented):
mostly looks good.  A few questions and one suggested function rename

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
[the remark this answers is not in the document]
ukstv: There is a [PR in the works](https://github.com/trufflesuite/ganache/pull/1135) for Ganache to support 1559, so this thing is not for long.

And, no, there is no way I can imagine to wrap Ganache into something with 1559 support.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -89,14 +95,24 @@ export default class EthereumBlockchainService implements BlockchainService {
...
+      const prevMaxPriorityFeePerGas = BigNumber.from(txData.maxPriorityFeePerGas || 0)
+      const nextMaxPriorityFeePerGas = EthereumBlockchainService.increaseGasPricePerAttempt(
+        feeData,
         attempt,
-        txData.gasPrice
+        prevMaxPriorityFeePerGas
+      )
+      const difference = nextMaxPriorityFeePerGas.sub(prevMaxPriorityFeePerGas)
+      if (feeData.maxPriorityFeePerGas) {
+        txData.maxFeePerGas = feeData.maxFeePerGas.add(difference)
```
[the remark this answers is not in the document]
ukstv: By definition, `maxFeePerGas = baseFee + maxPriorityFeePerGas`. If we increase `maxPriorityFee`, we should make sure the equality still stands. If not adjusting for the difference we might end up `maxFeePerGas` being less than `baseFee + maxPriorityFeePerGas` (with maxPriorityFeePerGas being fixed). So this leads to `baseFee` being set too low, which reduces the chance of a tx being mined regardless of our `maxPriority` increase.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -109,35 +125,31 @@ export default class EthereumBlockchainService implements BlockchainService {
    * previous attempt's gas, even if the gas price on chain has gone down since then. This is because
    * retries of the same transaction (using the same nonce) need to have a gas price at least 10% higher
    * than any previous attempts or the transaction will fail.
-   * @param currentGasEstimate
+   * @param feeData
    * @param attempt
    * @param previousGas
    */
   static increaseGasPricePerAttempt(
```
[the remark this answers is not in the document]
ukstv: Not quite, we increase the full gas price here, including `maxFeePerGas`.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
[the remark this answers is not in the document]
stbrody: Can we move this logic into `increaseGasPricePerAttempt` then?  Feels like half the logic happens there and half of it happens here.  Makes it confusing to follow

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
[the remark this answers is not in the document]
stbrody: feels like it wouldn't be that hard to mock the `getFeeData()` and `sendTransaction()` functions to coerce the txData into the right format?  i.e. making `getFeeData()` replace the `gasPrice` field with the `maxPriorityFeePerGas` field, and `sendTransaction()` replace the `maxPriorityFeePerGas` field with a `gasPrice` field.  Basically just doing the same thing as we are doing here and in `increaseGasPricePerAttempt` to accommodate non-1559 transactions, but moved into test code that injects a mock that intercepts the messages before passing them on?  That would put all the non-1559 support code squarely within the test file and allow the production code to only worry about the formats for 1559 transactions.

If there's something that makes this harder than I'm realizing, then at the very least we should add TODO comments to mark every line in the production code that only exists to support Ganache so it'll be clear when we can rip them out.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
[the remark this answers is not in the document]
stbrody: Then can we update the comments for this function at least?  It really doesn't seem likely to be immediately obvious to new readers what this is doing and the relationship between the `maxFeePerGas`, `maxPriorityFeePerGas`, and the math that's being done within this method

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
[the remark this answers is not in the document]
stbrody: or maybe `increaseGasPricePerAttempt` should just return the difference (and possibly be renamed), then it could be added to both `maxPriorityFeePerGas` and `maxFeePerGas` together here?  I just want to make sure it's easy to understand what this function is doing and how to use it correctly

stbrody (changes requested):
I still think we need to do more to make it clear how to use `increaseGasPricePerAttempt` correctly.  I'm going to add @v-stickykeys  as a reviewer, as she reviewed the original PR that added the retry logic with increased fees, and had helpful comments back then on making sure the logic was clear and understandable.

I also think it should be possible to move the non1559 txn support into test-only code, but would be okay with just labeling which code exists solely for compatibility with Ganache.  We could make a ticket to remove the backwards compatibility-specific code once Ganache supports 1559, and add comments like `TODO(<ticket number>): remove once Ganache supports EIP1559 transactions` on the relevant lines of code.

src/services/blockchain/__tests__/eth-bc-service.test.ts
```diff
@@ -81,16 +82,24 @@ describe('ETH service connected to ganache', () => {
   })
 
   test('gas price increase math', () => {
-    const gasEstimate = BigNumber.from(1000)
+    const gasEstimate: providers.FeeData = {
+      maxFeePerGas: BigNumber.from(2000),
```
v-stickykeys, line 86: What's the difference between "maxFeePerGas" and "gasPrice"?

src/services/blockchain/__tests__/eth-bc-service.test.ts
```diff
@@ -247,14 +258,12 @@ describe('ETH service with mock wallet', () => {
...
-    const [txData0, attemptNum0, network0] = mockTrySendTransaction.mock.calls[0]
+    const [txData0, attemptNum0] = mockTrySendTransaction.mock.calls[0]
     expect(attemptNum0).toEqual(0)
-    expect(network0).toEqual(config.blockchain.connectors.ethereum.network)
     expect(txData0).toMatchSnapshot()
 
-    const [txData1, attemptNum1, network1] = mockTrySendTransaction.mock.calls[1]
+    const [txData1, attemptNum1] = mockTrySendTransaction.mock.calls[1]
     expect(attemptNum1).toEqual(1)
-    expect(network1).toEqual(config.blockchain.connectors.ethereum.network)
```
v-stickykeys, line 257: Why don't we use network anymore?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -25,17 +25,22 @@ const POLLING_INTERVAL = 15 * 1000 // every 15 seconds
...
   public static make(config: Config): EthereumBlockchainService {
     const { network } = config.blockchain.connectors.ethereum
     const { host, port, url } = config.blockchain.connectors.ethereum.rpc
 
     let provider
-    if (url && url != '') {
+    if (url) {
       provider = new ethers.providers.JsonRpcProvider(url)
-    } else if (host && host != '' && port && port != '') {
+    } else if (host && port) {
```
v-stickykeys, line 43: Why not check for empty strings?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -83,20 +89,33 @@ export default class EthereumBlockchainService implements BlockchainService {
...
+      )
+      const difference = nextMaxPriorityFeePerGas.sub(prevMaxPriorityFeePerGas)
+      if (feeData.maxPriorityFeePerGas) {
+        txData.maxFeePerGas = feeData.maxFeePerGas.add(difference)
+        txData.maxPriorityFeePerGas = nextMaxPriorityFeePerGas
+      } else {
+        txData.gasPrice = nextMaxPriorityFeePerGas
+      }
+      logger.debug(
+        'Estimated maxPriorityFeePerGas (in wei): ' + nextMaxPriorityFeePerGas.toString()
```
v-stickykeys, line 117: Seems like we should log the total fee not just the priority fee

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
v-stickykeys: Disagree. For debugging it is helpful to have ganache support. But maybe in that case we would override?

However still think backwards compatibility should be supported in general

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -109,35 +128,31 @@ export default class EthereumBlockchainService implements BlockchainService {
...
-    currentGasEstimate: BigNumber,
+    feeData: providers.FeeData,
     attempt: number,
     previousGas: BigNumberish | undefined
   ): BigNumber {
-    const tenPercent = currentGasEstimate.div(10)
-    const additionalGas = tenPercent.mul(attempt)
-    const newGas = currentGasEstimate.add(additionalGas)
+    // Ganache still does not support 1559, so here is a workaround
+    const maxPriorityFeePerGas = feeData.maxPriorityFeePerGas || feeData.gasPrice
```
v-stickykeys, line 141: Why is it or gasprice? Is this the workaround? If so it seems better to have two helper methods for the calculation. One that is like legacy and one that is 1559

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -109,35 +128,31 @@ export default class EthereumBlockchainService implements BlockchainService {
...
+    // Ethers BigNumber does not support non-integer numbers for operations
+    const decimalPart = maxPriorityFeePerGas.div(10)
```
v-stickykeys, line 143: What does decimal part mean?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -109,35 +128,31 @@ export default class EthereumBlockchainService implements BlockchainService {
...
+    const increase = decimalPart.mul(attempt)
+    const newGas = maxPriorityFeePerGas.add(increase)
```
v-stickykeys, line 145: What is new gas? Is it the total fee we will pay per gas?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -269,91 +271,93 @@ export default class EthereumBlockchainService implements BlockchainService {
...
+            await Utils.delay(5000)
+          }
         }
       }
-    }
+    })
+  }
+
+  async withWalletBalance<T>(action: (balance: BigNumber) => Promise<T>): Promise<T> {
+    const before = await this.wallet.provider.getBalance(this.wallet.address)
```
v-stickykeys, line 347:
```suggestion
    const startingWalletBalance = await this.wallet.provider.getBalance(this.wallet.address)
```

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -269,91 +271,93 @@ export default class EthereumBlockchainService implements BlockchainService {
...
+    logMetric.ethereum({
+      type: 'walletBalance',
+      balance: ethers.utils.formatUnits(before, 'gwei'),
+    })
+    logger.imp(`Current wallet balance is ` + before)
+
+    const result = await action(before)
 
-    const finalWalletBalance = await this.wallet.provider.getBalance(this.wallet.address)
+    const after = await this.wallet.provider.getBalance(this.wallet.address)
```
v-stickykeys, line 356:
```suggestion
    const endingWalletBalance = await this.wallet.provider.getBalance(this.wallet.address)
```

v-stickykeys (changes requested):
Big PR 🙃

I'm not fully following the fee setting yet. Would like to block this until we can get it a lot easier to read. Added a bunch of questions which I think should have more obvious answers in the code. Probably a combination of variable naming and helper methods. Will provide more suggestions shortly

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stbrody: Fair. In that case then I agree with you that we should do more to separate out the logic for 1559 transactions from the logic for legacy style transactions so it's clear what use case each piece of code is used for.

PR #498, opened 2021-09-08
fix(ci): replace = with colons for set output

v-stickykeys:
Env var was coming up as blank bc it wasn't being set correctly


smrz2001 (approved):
:shipit:

PR #506, opened 2021-10-20
feat: Support EIP1559 ethereum transactions

ukstv:
- update ethers to the latest one
- use `provider.getFeeData` to get estimated gas cost: 1559 `maxFeePerGas` and `maxPriorityFeePerGas`, and vanilla `gasPrice`.

FYI. EIP1559 now uses two components of gas cost: `baseFee` and `maxPriorityFeePerGas`. `maxPriorityFeePerGas` is a tip to a miner to include a transaction into a block. `baseFee` is a slowly changing amount of gas or ether that is going to be burned. `baseFee` is set by _network_. Since we do not know what `baseFee` will be, EIP1559 introduces `maxFeePerGas` which is an absolute maximum you are willing to pay for a transaction. `maxFeePerGas` must be `>= maxPriorityFeePerGas + baseFee`. The inequality here is to accommodate for changes in `baseFee`. If `maxFeePerGas` appears to be less than the sum, the transaction is underpriced. If it is greater than the sum (`maxFeePerGas = maxPriorityFeePerGas + baseFee + δ`):
- if `baseFee` changes up to `δ`, the transaction can be mined still; `δ` is like a safety buffer;
- transaction fee that is deducted from your wallet still equals `maxPriorityFeePerGas + baseFee`, no matter what `maxFeePerGas` you have set.

To price a 1559 transaction, we use an estimate from `provider.getFeeData`. It returns `maxFeePerGas` and `maxPriorityFeePerGas`. It is worth noting here that `maxFeePerGas` returned from ethers uses a [widely recommended](https://www.blocknative.com/blog/eip-1559-fees) formula: `(baseFee of the latest block)*2 + (constant 2.5Gwei)`. If we only increase `maxPriorityFeePerGas` per attempt, we effectively deduct from our baseFee safety buffer `δ` which reduces transaction's chances. Our intent though is to increase a transaction's "mineability". So, when increasing `maxPriorityFeePerGas` we also increase `maxFeePerGas` by the same amount. Now the safety buffer reflects current network conditions, and we actually increase our transaction's "mineability".


stbrody (commented):
Lots of really great cleanups here.  A bunch of little comments throughout, but overall I think this is looking really good

stbrody (commented):
this lgtm whenever Val is happy, thanks!

v-stickykeys (commented):
👍

v-stickykeys (approved):
Good with me! Can make some small cleanups as mentioned in comments.

PR #512, opened 2021-11-10
feat: Recalculate maxFeePerGas via baseFee

ukstv:
Use more straight-forward calculation of maxFeePerGas. It does not depend now on previous transaction fee.

### DIFF
--- a/src/services/blockchain/__tests__/eth-bc-service.test.ts
+++ b/src/services/blockchain/__tests__/eth-bc-service.test.ts
@@ -13,6 +13,7 @@ import EthereumBlockchainService, { MAX_RETRIES } from '../ethereum/ethereum-blo
 import { BigNumber } from 'ethers'
 import type { FeeData } from '@ethersproject/abstract-provider'
 import { ErrorCode } from '@ethersproject/logger'
+import type { TransactionRequest } from '@ethersproject/abstract-provider'
 
 describe('ETH service connected to ganache', () => {
   jest.setTimeout(25000)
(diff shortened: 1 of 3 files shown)


stbrody (commented):
the fact that this change doesn't break any of the tests suggests to me that we're missing test coverage.  Can you add a test case that would reproduce the issue we saw on clay where we underpriced a transaction?

ukstv:
Added the tests. They fail indeed, if EthereumBlockchainService logic changes.

src/services/blockchain/__tests__/eth-bc-service.test.ts
```diff
@@ -93,11 +98,63 @@ describe('ETH service connected to ganache', () => {
...
+
+  test('legacy transaction', async () => {
+    const legacyProvider = Object.assign({}, provider, {
+      getFeeData: jest.fn(() => ({ gasPrice: feeData.gasPrice })),
+    })
+    const ethBc = await buildBlockchainService(legacyProvider)
+    const txData = await ethBc._buildTransactionRequest(cid)
+    for (const attempt of [0, 1, 2]) {
+      await ethBc.setGasPrice(txData, attempt)
+      expect(txData).toMatchSnapshot()
```
stbrody, line 144: sigh.  this test doesn't really accomplish what I was hoping for.  snapshot tests should be used very sparingly IMO as they don't really convey any semantic value to the reader about what they are trying to test.  They'll tell us if the response changes, but not in a way that lets future people making changes to this code understand whether that change is good or bad.  I was hoping for a test that specifically targeted reproducing the issue that we saw in our environment where some anchors were failing due to being underpriced on a transaction retry because the gas price had fallen but transaction retries always have to be priced higher than previous attempts.  That way we would be specifically testing the condition that lead to that underpricing happening, and have a test that is clearly testing for transaction underpricing due to retry logic and changing gas fees.  That way even if the exact gas pricing logic changes in the future, we'd still have a test case that forces the author of the new gas price logic (who might not be you or I) to remember to consider this scenario.

In any case, it's not important enough to keep holding up this review any longer, but something to keep in mind for the future.

stbrody (approved):

PR #515, opened 2021-11-29
fix: Only mark selected requests as processing

stbrody:
Currently the CAS anchor process starts by loading *every* pending request from the DB into memory, and marking all of them as PROCESSING.  Then it works to build a batch out of 2^merkleDepthLimit streams (which is usually 10, so 1024 streams).  Whatever requests it doesn't wind up selecting in the batch wind up getting returned to PENDING.

This patch makes 2 changes:
* first, it limits the number of requests from the DB pulled into memory to twice the number of streams we want based on the merkleDepthLimit.  In practice this will be a max of 2048k requests that we'll consider for any given batch.  If those 2048 requests don't turn out to include 1024 unique, loadable streams, then we proceed with an under-full batch, rather than going through the complexity of going back to the database to fetch more requests in an attempt to fill the batch.
* second, it doesn't mark requests as PROCESSING until after we have successfully loaded the corresponding streams from Ceramic and selected those requests for inclusion in the batch.  This has the downside that if we were to try to run two CAS instances at the same time on the same database, now we would wind up processing the same requests in each instance.  But since we already avoid running multiple CAS instances at once, this is deemed acceptable for now to simplify the logic and prevent us from moving a whole bunch of requests in to PROCESSING only to be sent back to PENDING shortly thereafter if they don't make it into the batch.

### DIFF
--- a/src/repositories/request-repository.ts
+++ b/src/repositories/request-repository.ts
@@ -93,7 +93,7 @@ export default class RequestRepository extends Repository<Request> {
   /**
    * Gets all requests by status
    */
-  public async findNextToProcess(): Promise<Request[]> {
+  public async findNextToProcess(limit: number): Promise<Request[]> {
     const now: number = new Date().getTime()
     const deadlineDate = new Date(now - this.config.expirationPeriod)
 
@@ -106,6 +106,7 @@ export default class RequestRepository extends Repository<Request> {
         processingStatus: RequestStatus.PROCESSING,
         deadlineDate: deadlineDate.toISOString(),
       })
+      .limit(limit)
       .getMany()
   }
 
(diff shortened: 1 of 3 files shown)


oed (approved):
lgtm
> second, it doesn't mark requests as PROCESSING until after we have successfully loaded the corresponding streams from Ceramic and selected those requests for inclusion in the batch. This has the downside that if we were to try to run two CAS instances at the same time on the same database, now we would wind up processing the same requests in each instance. But since we already avoid running multiple CAS instances at once, this is deemed acceptable for now to simplify the logic and prevent us from moving a whole bunch of requests in to PROCESSING only to be sent back to PENDING shortly thereafter if they don't make it into the batch.

This seems completely fine as we are very far from needing to do parallel CAS runs (we would instead do them more frequently than twice a day).

PR #539, opened 2021-12-13
chore: Update batch summary log message

stbrody:


src/services/anchor-service.ts
```diff
@@ -171,6 +162,19 @@ export default class AnchorService {
     }
     logger.imp(`Service successfully anchored ${anchors.length} CIDs.`)
 
+    logEvent.anchor({
+      type: 'anchorRequests',
```
stbrody, line 166: we should be able to build a bunch of good dashboards off this message now, and can update the saved search you already made to include much more detailed information about exactly what happened in the batch.

v-stickykeys (approved):
LGTM

PR #541, opened 2021-12-14
feat: Publish anchor commits on multiple streams concurrently

stbrody:


oed (approved):
lgtm

stbrody:
Thanks for a review Joel but I'd still like @ukstv to take a look at this and https://github.com/ceramicnetwork/ceramic-anchor-service/pull/540 to double check me on the concurrency control

ukstv (approved):

stbrody:
We are keeping up without this.  Closing this for now to avoid potentially destabilizing anything or causing more anchors to fail due to extra load.  If we need to speed up anchoring in the future we can re-consider this

PR #549, opened 2021-12-15
feat: Load stream first, then only send multiquery for missing commits

stbrody:


src/services/anchor-service.ts
```diff
@@ -564,8 +564,36 @@ export default class AnchorService {
...
+      candidate.setTipToAnchor(stream)
+      return
+    }
+
+    // If there were CIDs that we have requests for but didn't show up in the stream state that
+    // we loaded from Ceramic, we can't tell if that is because those commits were rejected by
+    // Ceramic's conflict resolution, or if our local Ceramic node just never heard about those
+    // commits before.  So we build a multiquery including all missing commits and send that to
+    // Ceramic, forcing it to at least consider every CID that we have a request for.
+    const queries = missingRequests.map((request) => {
```
oed, line 596: Can we add a log here so that we can see if there are missing requests present? I suspect we want to see this info in the future.

stbrody: If the commit is still missing after the multiquery we log it.  You want to also log the commits that are missing from the initial load but might still show up after the multiquery?  Do you want each individual CID logged, or would just a count of how many are missing be enough?

oed: Yes. A number might be enough. What do you think?

oed: If the CIDs are included it would be easier to debug though.

stbrody: okay done

oed (approved):
lgtm

PR #550, opened 2021-12-16
feat: Create anchor commits directly on ipfs

stbrody:


v-stickykeys (approved):
LGTM

v-stickykeys (approved):
👍

src/services/anchor-service.ts
```diff
@@ -306,10 +306,12 @@ export default class AnchorService {
...
-      const anchorCid = await this.ceramicService.publishAnchorCommit(
-        candidate.streamId,
-        ipfsAnchorCommit
-      )
+      // TODO(#548): Publish anchor commits via Ceramic
+      // const anchorCid = await this.ceramicService.publishAnchorCommit(
+      //   candidate.streamId,
+      //   ipfsAnchorCommit
+      // )
+      const anchorCid = await this.ipfsService.storeRecord(ipfsAnchorCommit)
```
oed, line 314: Why do we need to store the anchor commit? This commit would be created by js-ceramic nodes when they get the anchored successful response?

oed (approved):
Not sure why we need to store the anchor commit itself into ipfs, but I guess it doesn't really hurt.

src/services/anchor-service.ts
stbrody: The CAS just passes the anchor commit CID to the ceramic node that requested the anchor. The ceramic node then loads the anchor commit from ipfs, applies it, and publishes it to pubsub.

If the commit doesn't exist on ipfs then loading the commit on ceramic will fail

oed: Oh I see.

PR #554, opened 2021-12-17
fix: don't ever pull in newer requests over older ones

stbrody:
this will make batches far more likely to be under-full, but avoids ever anchoring requests out of order


smrz2001 (approved):
🚀

PR #558, opened 2021-12-17
chore: Consider a much larger number of requests per batch

stbrody:


src/services/anchor-service.ts
```diff
@@ -544,10 +550,11 @@ export default class AnchorService {
...
       failedRequests.push(...candidate.failedRequests)
       conflictingRequests.push(...candidate.rejectedRequests)
+    }
 
-      if (numSelectedCandidates >= candidateLimit) {
-        break
-      }
+    for (let i = numSelectedCandidates; i < candidates.length; i++) {
+      const unselectedCandidate = candidates[i]
+      unprocessedRequests.push(...unselectedCandidate.requests)
```
stbrody, line 557: this will cause the number of unprocessedRequests to be logged later on, which will give us pretty good visibility into the size of our backlog in each batch (so long as the backlog is less than 100 * 1024 requests large)

oed (approved):

PR #560, opened 2021-12-18
chore: Do not log every single request ID in batch summary message

stbrody:
It's overly verbose, and now that we pull in all pending requests, the log line has gotten so long it seems to have passed some maximum in Log Insights - Log insights is breaking the line up into multiple lines, which is breaking some of the saved queries we use to get reports on what the batch did


stbrody:
I'm going to go ahead and merge this straight away.  @v-stickykeys, when you're back please take a look and let me know if those requestsIds in that log message are used anywhere I'm not aware of.

PR #562, opened 2021-12-20
feat: Only anchor if batch is at least half full

stbrody:


src/services/anchor-service.ts
```diff
@@ -453,6 +453,13 @@ export default class AnchorService {
     logger.debug(`Grouping requests by stream`)
     const candidates = AnchorService._buildCandidates(requests)
 
+    if (candidates.length < candidateLimit / 2) {
```
stbrody, line 456: this is the core change

src/services/anchor-service.ts
```diff
@@ -530,9 +537,15 @@ export default class AnchorService {
       candidateLimit = candidates.length
     }
 
-    for (let i = 0; i < candidateLimit; i++) {
+    for (let i = 0; i < candidates.length; i++) {
       const candidate = candidates[i]
 
+      if (numSelectedCandidates >= candidateLimit) {
```
stbrody, line 543: this is an unrelated bug fix I snuck in here.  I can separate it into its own commit if you think it necessary, just did this for expediency.  But the old logic to select unprocessedRequests after the first loop had an error in the logic of where it started processing candidates that causes it to report the wrong number of unprocessedRequests

oed (approved):
lgtm. I wonder if the process runs for a really short time if it's aborted we will spin up and down a lot of processes all the time?

stbrody:
> lgtm. I wonder if the process runs for a really short time if it's aborted we will spin up and down a lot of processes all the time?

yeah good point.  Maybe I'll make it sleep for a while before shutting down if there's not enough to anchor.  30 minutes?  An hour?

stbrody:
I realized that a fixed minimum of 1/2 the desired number of streams (so 512 for us in all our envs) would break our integration tests on dev which rely on small anchor batches.  So the minimum needs to be configurable so we can set it to 1 on dev, but 512 on clay and prod.  PTAL @oed

src/services/anchor-service.ts
```diff
@@ -128,9 +128,21 @@ export default class AnchorService {
       // max depth of the merkle tree.
       streamCountLimit = Math.pow(2, this.config.merkleDepthLimit)
     }
-    const [candidates, groupedRequests] = await this._findCandidates(requests, streamCountLimit)
+    const minStreamCount = this.config.minStreamCount
+      ? this.config.minStreamCount
+      : streamCountLimit / 2
```
oed, line 133:
```suggestion
    const minStreamCount = this.config.minStreamCount || streamCountLimit / 2
```

src/services/anchor-service.ts
```diff
@@ -128,9 +128,21 @@ export default class AnchorService {
...
+    const [candidates, groupedRequests] = await this._findCandidates(
+      requests,
+      streamCountLimit,
+      minStreamCount
+    )
     if (candidates.length === 0) {
-      logger.debug('No candidates found. Skipping anchor.')
+      logger.imp('No candidates found. Skipping anchor.')
+      logger.debug(
+        'Sleeping 30 minutes before shutting down to prevent constantly running empty anchor batches'
```
oed, line 142: Maybe just 10 minutes in case request count goes up a lot?

oed (approved):

PR #564, opened 2021-12-27
fix: remove anchor delay for dev

v-stickykeys:
# Remove anchor delay for dev

## Description

Don't sleep for 10 minutes when anchoring on dev.

## Notes
- Requires anchor task definition on infra to have `NODE_ENV` set to `dev`
- Should remove the sleep entirely and keep the control of when anchors occur outside of the CAS source code

### DIFF
--- a/src/services/anchor-service.ts
+++ b/src/services/anchor-service.ts
@@ -141,13 +141,15 @@ export default class AnchorService {
       streamCountLimit,
       minStreamCount
     )
-    if (candidates.length === 0 && process.env.NODE_ENV !== 'dev') {
+    if (candidates.length === 0) {
       logger.imp('No candidates found. Skipping anchor.')
-      logger.debug(
-        'Sleeping 10 minutes before shutting down to prevent constantly running empty anchor batches'
-      )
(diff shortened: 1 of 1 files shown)


stbrody:
why do this instead of just setting `MIN_STREAM_COUNT` to 1 like we discussed on discord?

stbrody:
I see, the problem is if you have 0 requests you wind up sleeping, which could throw off timing of integration tests.

lgtm

src/services/anchor-service.ts
```diff
@@ -141,7 +141,7 @@ export default class AnchorService {
       streamCountLimit,
       minStreamCount
     )
-    if (candidates.length === 0) {
+    if (candidates.length === 0 && process.env.NODE_ENV !== 'dev') {
```
stbrody, line 144: we only want to skip the sleep if we're in dev, not the early return

stbrody (changes requested):
actually, need to update the code change slightly so we still early return without anchoring if there's no pending requests, we just don't want to sleep 10 minutes in that case on dev

stbrody (approved):
lgtm

PR #570, opened 2022-01-09
hotfix: lock colors dependency to stop infinite loop

v-stickykeys:
See https://github.com/ceramicnetwork/js-ceramic/pull/1981 for details--addresses the same issue.

TLDR; circumvents infinite loop in logger


oed (approved):

oed:
Does this need any sort of release?

PR #577, opened 2022-02-02
chore: don't mark candidate requests as processing

stephhuynh18:
# Don't mark candidate requests as processing

## Description

I did not remove the processing status as it may be used in the future when we decide to run CAS instances in parallel. 

## How Has This Been Tested?

Tests are running


src/services/anchor-service.ts
```diff
@@ -488,11 +488,10 @@ export class AnchorService {
       for (const candidate of candidates) {
         groupedRequests.acceptedRequests.push(...candidate.acceptedRequests)
       }
-      logger.debug(
-        `Marking ${groupedRequests.acceptedRequests.length} pending requests as processing`
-      )
+
+      // set candidates that will be anchored as pinned to be garbage collected later
       await this.requestRepository.updateRequests(
```
stbrody, line 493: I think we should remove this update entirely.  It does mean that if we crash some streams may be pinned that aren't marked that way in the database, but that's already the case today, this just makes that window slightly larger.  But in the vast majority of cases, those streams will be picked up in the next batch and if that batch finishes successfully then the state will be correct.

This update slows things down and floods the logs, for very little real value.

stbrody (changes requested):
Can we rename the `PROCESSING` status to `DEPRECATED_PROCESSING`?
 https://github.com/ceramicnetwork/ceramic-anchor-service/blob/b1977d8555860527eaf43e3411fbfa0661207848/src/models/request-status.ts#L3

stbrody:
We should also update `findNextToProcess` to remove the special casing of the PROCESSING status: https://github.com/ceramicnetwork/ceramic-anchor-service/blob/b1977d8555860527eaf43e3411fbfa0661207848/src/repositories/request-repository.ts#L96

and update the README to remove references to PROCESSING.

stephhuynh18:
@stbrody yup sounds good. I wasn't sure if we just wanted to "pause" the status or deprecate it. I'll have those changes in shortly.

src/repositories/request-repository.ts
```diff
@@ -98,10 +98,6 @@ export class RequestRepository extends Repository<Request> {
       .createQueryBuilder('request')
```
stbrody, line 98: can remove `now` and `deadlineDate` above now as they are no longer used

stbrody (approved):
LGTM, but let's hold off on merging this until the CAS is working on dev again to avoid destabilizing any further

PR #590, opened 2022-02-16
feat: Add version to bloom filter metadata

stephhuynh18:
## How Has This Been Tested?

Added semver check to tests

### DIFF
--- a/src/merkle/merkle-objects.ts
+++ b/src/merkle/merkle-objects.ts
@@ -12,12 +12,12 @@ import { BloomFilter } from 'bloom-filters'
 import { StreamID } from '@ceramicnetwork/streamid'
 
 const packageJson = JSON.parse(
-  fs.readFileSync(new URL('../../package.json', import.meta.url), 'utf8')
+  fs.readFileSync(new URL('../../node_modules/bloom-filters/package.json', import.meta.url), 'utf8')
 )
 
 const BLOOM_FILTER_TYPE = 'jsnpm_bloom-filters'
 const BLOOM_FILTER_FALSE_POSITIVE_RATE = 0.0001
-const BLOOM_FILTER_VERSION = packageJson['dependencies']['bloom-filters']
+const bloomFilterVersion = packageJson['version']
 
 export interface CIDHolder {
   cid: CID
@@ -258,7 +258,7 @@ export class BloomMetadata implements MetadataFunction<Candidate, TreeMetadata> 
       bloomFilter: {
         type: BLOOM_FILTER_TYPE,
         data: serializedBloomFilter,
-        version: BLOOM_FILTER_VERSION,
+        version: bloomFilterVersion,
       },
     }
   }
--- a/src/merkle/merkle-objects.ts
+++ b/src/merkle/merkle-objects.ts
@@ -256,9 +256,8 @@ export class BloomMetadata implements MetadataFunction<Candidate, TreeMetadata> 
     return {
       numEntries: leaves.length,
       bloomFilter: {
-        type: BLOOM_FILTER_TYPE,
+        type: `${BLOOM_FILTER_TYPE}-v${bloomFilterVersion}`,
         data: serializedBloomFilter,
-        version: bloomFilterVersion,
       },
     }
   }
--- a/src/merkle/merkle-objects.ts
+++ b/src/merkle/merkle-objects.ts
@@ -1,4 +1,5 @@
 import { CID } from 'multiformats/cid'
+import * as fs from 'fs'
 
 import { AnchorStatus, Stream, StreamMetadata } from '@ceramicnetwork/common'
 import { CompareFunction, MergeFunction, MetadataFunction, Node, TreeMetadata } from './merkle.js'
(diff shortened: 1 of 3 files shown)


src/merkle/merkle-objects.ts
```diff
@@ -10,8 +11,13 @@ import { IpfsService } from '../services/ipfs-service.js'
 import { BloomFilter } from 'bloom-filters'
 import { StreamID } from '@ceramicnetwork/streamid'
 
+const packageJson = JSON.parse(
+  fs.readFileSync(new URL('../../package.json', import.meta.url), 'utf8')
+)
+
 const BLOOM_FILTER_TYPE = 'jsnpm_bloom-filters'
 const BLOOM_FILTER_FALSE_POSITIVE_RATE = 0.0001
+const BLOOM_FILTER_VERSION = packageJson['dependencies']['bloom-filters']
```
stbrody, line 20: unfortunately this approach will only tell us the version that is specified in the package.json, not what was actually installed by npm, which could be different.  Given that the bloom-filters package literally had a breaking change in a patch version release (which is what started all this), we need to record the *actual* version being used.  I don't actually know if there's an easy way to figure that out at runtime in javascript, maybe @ukstv has an idea

stbrody, line 20: nitpick: All CAPS variable names should be reserved for hard-coded constants, not anything that could be different at runtime across invocations.

stbrody (changes requested):

src/merkle/merkle-objects.ts
stephhuynh18: AHHH i see i would have hoped that they respected semver hahaha

stephhuynh18: A similar to solution to what I have here is using `node_modules/bloom-fiters/package.json` instead of our `package.json` dependencies. This solution will provide the exact version installed. I am not sure if this is the best solution so i'll push the code and see what @ukstv says.

stephhuynh18:
Not sure if this was ever approved so I will request reviews again

stbrody:
was there ever a change to fix how we check the version to make sure we see the version actually installed not just what's specified in the package.json?

stephhuynh18:
:o oops I forgot to push =3

src/merkle/merkle-objects.ts
```diff
@@ -10,8 +11,13 @@ import { IpfsService } from '../services/ipfs-service.js'
 import { BloomFilter } from 'bloom-filters'
 import { StreamID } from '@ceramicnetwork/streamid'
 
+const packageJson = JSON.parse(
+  fs.readFileSync(new URL('../../node_modules/bloom-filters/package.json', import.meta.url), 'utf8')
```
stbrody, line 15: this seems reasonable to me but I'd like @ukstv to take a look and make sure this is robust enough and will work in all the different environments in which anyone might want to run a CAS.  For example will this work on Windows?  Is anything different required after a move to ES2020/ESM?

src/merkle/merkle-objects.ts
```diff
@@ -249,7 +255,11 @@ export class BloomMetadata implements MetadataFunction<Candidate, TreeMetadata>
     const serializedBloomFilter = bloomFilter.saveAsJSON()
     return {
       numEntries: leaves.length,
-      bloomFilter: { type: BLOOM_FILTER_TYPE, data: serializedBloomFilter },
+      bloomFilter: {
+        type: BLOOM_FILTER_TYPE,
+        data: serializedBloomFilter,
+        version: bloomFilterVersion,
```
ukstv, line 261: Do we parse `type` anywhere? IMHO [soft suggestion], we should have version as part of `type` field here. On the parsing side it would be easier to have a pattern matching mechanism for just single `type` field.
New `type` would look like `${BLOOM_FILTER_TYPE}-v${bloomFilterVersion}`.

src/merkle/merkle-objects.ts
ukstv: @stbrody Confirm. This thing will work on Windows.

ukstv (approved):
Would like @stbrody and @stephhuynh18 considering putting version information in the `type` field, but otherwise LGTM.

src/merkle/merkle-objects.ts
ukstv: I am okay with that. As we discussed with @stephhuynh18 , better way would be to have our own bloom-filter (or XOR filter) implementation, to reduce potential risks from external dependency mismanagement. But, it increases our scope, so I am all right with the current @stephhuynh18 approach.

src/merkle/merkle-objects.ts
ukstv: I mean, I did run similar code on Windows, and it worked.

src/merkle/merkle-objects.ts
stbrody: I don't feel strongly either way.  Generally I prefer structured objects to doing string parsing.  But in this case I could also see value in causing indexers to fail fast if they encounter a `type` value they've not seen before, rather than documenting that they need to also check `version` and trust that they actually read the docs and implemented version handling correctly

stbrody: I think I'm actually becoming convinced that doing as @ukstv suggested is the way to go.  I'm normally pretty opposed to stuffing too much information into a single string field when a more structured alternative is available, but in this case I do think it make sense.  The `type` field exists to inform users what bloom filter implementation they need to use to interpret the `data` field.  Given that breaking changes can happen to whatever bloom filter package we use, the combination of package name + version is really required to be able to correctly identify the implementation needed to correctly use `data`.  So if you cannot meaningfully use `type` without `version`, then it makes sense to combine them into one thing that is actually usable.

stbrody (approved):
lgtm

PR #614, opened 2022-04-06
Mock ipfs client instead of ipfs service in test

stephhuynh18:
Changed the anchor-service test to use a mocked ipfs client instead of a mocked ipfs service. This will allow us to get into more granular testing of the ipfs-service

### DIFF
--- a/src/services/__tests__/anchor-service.test.ts
+++ b/src/services/__tests__/anchor-service.test.ts
@@ -13,7 +13,7 @@ import { IpfsService } from '../ipfs-service.js'
 import { AnchorRepository } from '../../repositories/anchor-repository.js'
 import { config } from 'node-config-ts'
 import { CommitID, StreamID } from '@ceramicnetwork/streamid'
-import { MockCeramicService, MockIpfsService } from '../../test-utils.js'
+import { MockCeramicService, MockIpfsClient } from '../../test-utils.js'
 import type { Connection } from 'typeorm'
 import { CID } from 'multiformats/cid'
 import { Candidate } from '../../merkle/merkle-objects.js'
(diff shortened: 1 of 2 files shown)


src/test-utils.ts
```diff
@@ -8,11 +8,40 @@ import { StreamID, CommitID } from '@ceramicnetwork/streamid'
...
   return CID.create(1, dagCBOR.code, create(0x12, randomBytes(32)))
 }
 
+export class MockIpfsClient {
+  constructor() {
+    this.reset()
+  }
+
+  private _streams: Record<string, any> = {}
+  pubsub
```
ukstv, line 23:
```suggestion
  pubsub: any
```

src/test-utils.ts
```diff
@@ -8,11 +8,40 @@ import { StreamID, CommitID } from '@ceramicnetwork/streamid'
...
+  dag
```
ukstv, line 24:
```suggestion
  dag: any
```

src/test-utils.ts
```diff
@@ -8,11 +8,40 @@ import { StreamID, CommitID } from '@ceramicnetwork/streamid'
...
+      subscribe: jest.fn(() => Promise.resolve()),
+    }
+    this.dag = {
+      get: jest.fn((cid: CID) => {
+        return Promise.resolve({ value: this._streams[cid.toString()] })
+      }),
+      put: jest.fn(async (record: Record<string, unknown>) => {
+        const cid = await randomCID()
+        this._streams[cid.toString()] = record
+        return Promise.resolve(cid)
```
ukstv, line 37:
```suggestion
        return cid
```

src/services/__tests__/anchor-service.test.ts
```diff
@@ -69,25 +69,39 @@ function createStream(id: StreamID, log: CID[], anchorStatus: AnchorStatus = Anc
   }
 }
 
+const mockIpfsClient = new MockIpfsClient()
+jest.mock('ipfs-http-client', () => {
```
ukstv, line 73: @stephhuynh18 If this works, it means, the test runs in CJS mode, IMO, which is not right long term. FYI.

stbrody: instead of mocking this library to override the `create` function, we could make `IpfsService.init()` take an optional param of the IpfsClient to use, and just manually provide the MockIpfsClient in the unit test

stbrody (approved):
lgtm with one optional suggestion to avoid relying on CJS jest mocking

src/services/__tests__/anchor-service.test.ts
stephhuynh18: @ukstv I forgot to change this to `unstable_mockModule`. Does using this make the tests run in ESM mode?

ukstv: > Does using this make the tests run in ESM mode?

Based on how the tests are run, it seems they run in ESM mode already. I was just confused that the tests worked. In my experience, jest `mock` and `requireActual` do not work in ESM mode. If they do now, it is perfect.

stephhuynh18: @stbrody will do that if the jest mocking messes things up later =) hopefully its actually fixed

PR #618, opened 2022-04-12
feat: publish pubsub update message

stephhuynh18:


src/__tests__/ceramic_integration.test.ts
```diff
@@ -127,7 +131,10 @@ async function makeCAS(
   configCopy.blockchain.connectors.ethereum.network = 'ganache'
   configCopy.blockchain.connectors.ethereum.rpc.port = minConfig.ganachePort + ''
   const childContainer = container.createChildContainer()
```
stbrody, line 133: I think it'd be cleaner if you made `childContainer` an argument to the makeCAS function, instead of a return value.  You can make and store a handle to the container in the test setup function, then pass it into makeCAS

src/__tests__/ceramic_integration.test.ts
```diff
@@ -389,5 +400,47 @@ describe('Ceramic Integration Test', () => {
       },
       60 * 1000 * 2
     )
+
+    test('Anchor discovered through pubsub', async () => {
+      jest.setTimeout(60 * 1000 * 2)
+      // In ceramic the stream waits for a successful anchor by polling the request endpoint of the CAS.
+      // We alter the CAS' returned request anchor status so that it is always pending.
+      // The stream will then have to hear about the successful anchor through pubsub
```
stbrody, line 408:
```suggestion
      // The ceramic node will then have to hear about the successful anchor through pubsub
```

stbrody (commented):
a few nitpicks, but overall looking good!

stbrody (approved):
lgtm, nice work!

src/services/ipfs-service.ts
```diff
@@ -6,13 +6,17 @@ import { logger } from '../logger/index.js'
 import * as dagJose from 'dag-jose'
 import type { IPFS } from 'ipfs-core-types'
 import { inject, singleton } from 'tsyringe'
-import { toCID } from '@ceramicnetwork/common'
+import { AnchorCommit, toCID } from '@ceramicnetwork/common'
+import { serialize, MsgType } from '@ceramicnetwork/core/lib/pubsub/pubsub-message'
```
stephhuynh18, line 10: FYI: I will update this to use `'@ceramicnetwork/core/pubsub/pubsub-message'` which is exported from the release candidate version of `core`, but I am currently having trouble with the named exports from `core`

PR #623, opened 2022-04-26
chore: Update log message

stbrody:


Jovonni (approved):
A standard log. This led me down the path of the entire file:
https://github.com/ceramicnetwork/ceramic-anchor-service/blob/37a747c2dd3213cef865a797e8e72183748749e0/src/services/anchor-service.ts

I now can see how we are building the merkle tree, and 

```typescript
async _createIPFSProof(tx: Transaction, merkleRootCid: CID): Promise<CID> {
```
```typescript
public async anchorRequests(): Promise<void> {
```

PR #644, opened 2022-05-31
feat: atomic request batches

stephhuynh18:
* Created a function that gathers requests for up to streamLimit streams and marks them as READY
* Undeprecated PROCESSING status

### DIFF
--- a/src/repositories/__tests__/request-repository.test.ts
+++ b/src/repositories/__tests__/request-repository.test.ts
@@ -14,7 +14,7 @@ import { RequestStatus } from '../../models/request-status.js'
 const MS_IN_MINUTE = 1000 * 60
 const MS_IN_HOUR = MS_IN_MINUTE * 60
 const MS_IN_DAY = MS_IN_HOUR * 24
-const MS_IN_MONTH = MS_IN_DAY * 31
+const MS_IN_MONTH = MS_IN_DAY * 30
 
 const generateRequests = async (override: Partial<Request>, count = 1): Promise<Request[]> => {
   const requests = await Promise.all(
--- a/src/repositories/__tests__/request-repository.test.ts
+++ b/src/repositories/__tests__/request-repository.test.ts
@@ -11,6 +11,34 @@ import { randomCID } from '../../test-utils.js'
 import { StreamID } from '@ceramicnetwork/streamid'
 import { RequestStatus } from '../../models/request-status.js'
 
+const MS_IN_MINUTE = 1000 * 60
+const MS_IN_HOUR = MS_IN_MINUTE * 60
+const MS_IN_DAY = MS_IN_HOUR * 24
+const MS_IN_MONTH = MS_IN_DAY * 31
+
+const generateRequests = async (override: Partial<Request>, count = 1): Promise<Request[]> => {
+  const requests = await Promise.all(
+    Array.from(Array(count)).map(async (_, i) => {
+      const request = new Request()
+      const cid = await randomCID()
+      request.cid = cid.toString()
+      request.streamId = new StreamID('tile', cid).toString()
+      request.status = RequestStatus.PENDING
+      request.createdAt = new Date(Date.now() - Math.random() * MS_IN_HOUR)
+      request.updatedAt = new Date(request.createdAt.getTime())
+
+      Object.assign(request, override)
+
+      const variance = Math.random() * 5
+      request.createdAt = new Date(request.createdAt.getTime() + MS_IN_MINUTE * (i + variance))
+      request.updatedAt = new Date(request.updatedAt.getTime() + MS_IN_MINUTE * (i + variance))
+      return request
+    })
+  )
+
+  return requests
+}
+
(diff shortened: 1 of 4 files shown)


src/repositories/request-repository.ts
```diff
@@ -145,4 +150,68 @@ export class RequestRepository extends Repository<Request> {
...
+      const earliestIsNotExpired =
+        streamsToAnchor.length > 0 && streamsToAnchor[0].createdAt > anchoringDeadline
+      if (earliestIsNotExpired && streamsToAnchor.length < streamLimit) {
+        return []
+      }
+
+      const streamIds = streamsToAnchor.map(({ streamId }) => streamId)
+
+      // retrieves all requests associated with the streams
+      const requests = await this.connection
```
stbrody, line 194: does this need to use `transactionalEntityManager`?

src/repositories/request-repository.ts
```diff
@@ -145,4 +150,68 @@ export class RequestRepository extends Repository<Request> {
...
+      const results = await this.updateRequests(
+        { status: RequestStatus.READY },
+        requests,
+        transactionalEntityManager
+      )
+
+      // if not all requests are updated
+      if (!results.affected || results.affected != requests.length) {
+        throw Error(
+          `A problem occured when updated requests to READY. Only ${requests.length}/${results.affected} requests were updated`
```
stbrody, line 210:
```suggestion
          `A problem occured when updated requests to READY. Only ${results.affected}/${requests.length} requests were updated`
```

stbrody: I think it's more common convention to show that only X out of an expected Y things happened as X/Y rather than Y/X

src/repositories/__tests__/request-repository.test.ts
```diff
@@ -131,4 +167,221 @@ describe('request repository test', () => {
...
+      await requestRepository.createRequests(requests)
+
+      const originaUpdateRequest = requestRepository.updateRequests
+      requestRepository.updateRequests = (fields, requests, manager) => {
+        throw new Error('test error')
+      }
+
+      try {
+        await requestRepository.findAndMarkReady(streamLimit).catch(() => {})
+        const requestsAfterUpdate = await getAllRequests(connection)
```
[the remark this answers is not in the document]
stephhuynh18: I have changed things to "expect that this function will throw". I don't think the other checks should be in the `finally` because the `finally` is just for cleanup.

src/repositories/request-repository.ts
```diff
@@ -15,6 +15,11 @@ import { inject, singleton } from 'tsyringe'
  * AnchorCommit available to the network.
  */
 const ANCHOR_DATA_RETENTION_WINDOW = 1000 * 60 * 60 * 24 * 30 // 30 days
+const MAX_ANCHORING_DELAY_MS = 1000 * 60 * 60 * 12 //12H
+
+const toComparisonDate = (date: Date) => {
```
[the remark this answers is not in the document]
stephhuynh18: yes removed

src/repositories/request-repository.ts
stephhuynh18: haha you are totally right

src/repositories/__tests__/request-repository.test.ts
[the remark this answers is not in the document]
stbrody: Oh, I had totally missed the "catch" chained onto the promise in the original version.  But yeah, explicitly expecting it to throw is even better

src/repositories/request-repository.ts
```diff
@@ -145,4 +147,68 @@ export class RequestRepository extends Repository<Request> {
...
+    return this.connection.transaction(isolationLevel, async (transactionalEntityManager) => {
+      // retrieves up to streamLimit unique streams with their earliest request createdAt value.
+      // this will only return streams associated with requests that are PENDING, or PROCESSING and needs to be retried
+      const streamsToAnchor = await transactionalEntityManager
+        .getRepository(Request)
+        .createQueryBuilder('request')
+        .select(['request.streamId', 'request.createdAt'])
+        .where('request.status = :processingStatus AND request.updatedAt < :retryDeadline', {
+          processingStatus: RequestStatus.PROCESSING,
+          retryDeadline: DateUtils.mixedDateToUtcDatetimeString(retryDeadline),
```
stbrody, line 173: any idea why this is necessary here but hasn't been necessary in any of the other queries we run that use timestamps?

stephhuynh18: this is where things get weird: I have this demo here https://github.com/ceramicnetwork/ceramic-anchor-service/blob/f1d17651dc1e266a4091fca2787a021dc503c657/src/repositories/__tests__/request-repository.test.ts#L414 that displays the different behaviours of the possible ways to do this date comparison. The only surefire way to retrieve a request by their date is by using this method. I have read sqlite does some conversion to the date based on timezones. I am not sure how to proceed HAHA.

PR #648, opened 2022-06-07
chore: refactor scheduler service

stephhuynh18:
* Uses rxjs
* does not use cron
* does not allow overlapping of task
* generic so it can take any task

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -3,7 +3,6 @@
   "port": 8081,
   "anchorControllerEnabled": false,
   "schedulerIntervalMS": 300000,
-  "cronExpression": "0/5 * * * ? *",
   "expirationPeriod": 5400000,
   "merkleDepthLimit": 0,
   "minStreamCount": 1,
--- a/config/default.json
+++ b/config/default.json
@@ -2,6 +2,7 @@
   "mode": "bundled",
   "port": 8081,
   "anchorControllerEnabled": false,
+  "schedulerIntervalMS": 300000,
   "cronExpression": "0/5 * * * ? *",
   "expirationPeriod": 5400000,
   "merkleDepthLimit": 0,
--- a/src/models/request-presentation.ts
+++ b/src/models/request-presentation.ts
@@ -1,6 +1,5 @@
 import { AnchorRepository } from '../repositories/anchor-repository.js'
 import { InvalidRequestStatusError, RequestStatus } from './request-status.js'
-import awsCronParser from 'aws-cron-parser'
 import { Request } from './request.js'
 
 /**
@@ -8,7 +7,7 @@ import { Request } from './request.js'
  */
 export class RequestPresentation {
   constructor(
-    private readonly cronExpression: string,
+    private readonly schedulerIntervalMS: number,
     private readonly anchorRepository: AnchorRepository
   ) {}
 
@@ -50,7 +49,6 @@ export class RequestPresentation {
         }
       }
       case RequestStatus.PENDING: {
-        const cron = awsCronParser.parse(this.cronExpression)
         return {
           id: request.id,
           status: RequestStatus[request.status],
@@ -60,7 +58,7 @@ export class RequestPresentation {
           message: request.message,
           createdAt: request.createdAt.getTime(),
           updatedAt: request.updatedAt.getTime(),
-          scheduledAt: awsCronParser.next(cron, new Date()),
+          scheduledAt: new Date(Date.now() + this.schedulerIntervalMS),
         }
       }
       case RequestStatus.PROCESSING:
--- a/src/services/__tests__/scheduler-service.test.ts
+++ b/src/services/__tests__/scheduler-service.test.ts
@@ -1,0 +1,64 @@
+import 'reflect-metadata'
+import { SchedulerService } from '../scheduler-service'
+import { config } from 'node-config-ts'
+import { container } from 'tsyringe'
+import { jest } from '@jest/globals'
+
+describe('scheduler service', () => {
+  jest.setTimeout(20000)
+  beforeAll(async () => {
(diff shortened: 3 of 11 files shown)


src/services/__tests__/scheduler-service.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        schedulerService.stop()
+        runChecks()
+      }
+
+      count = count + 1
+      return Promise.resolve()
+    })
+
+    schedulerService.start(task)
+    const runChecks = () => {
```
stbrody, line 30: is it safe to define this function after calling `scheduleService.start` with a task that may try to call it?

src/services/__tests__/scheduler-service.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      // the task runs once right at the start before running every X seconds
+      expect(task.mock.calls.length).toEqual(numberOfRunsBeforeDone + 1)
+      done()
+    }
```
stbrody, line 34: doesn't this test need to wait for the schedulerService to shut down?  otherwise couldn't the test finish while the scheduler is still running, before `runChecks` even gets called?

What happens if you put a purposefully failing assertion in runChecks?  Would be interesting to see how jest handles that assertion failing if the test has already completed running.

stbrody: You could change `schedulerService.start()` to return a promise that resolves when the service shuts down, and await that promise at the end of the test.

src/services/__tests__/scheduler-service.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    let count = 0
+    const task = jest.fn(() => {
+      if (count === numberOfRunsBeforeDone) {
+        schedulerService.stop()
+        runChecks()
+      }
+
+      count = count + 1
+
+      if (count > 3) {
```
stbrody, line 50: nitpick: inconsistent to use a constant for `numberOfRunsBeforeDone` but not for this threshold here.  Either use one for both or for neither

src/services/__tests__/scheduler-service.test.ts
stbrody: OH!  I see now the `done` argument.  I didn't realize that you could tell jest not to consider a test finished until you call a given function.  Cool.  Maybe then just add a comment as the very last line of the test `// test doesn't complete until 'done()' is called` so it's clearer to future readers.

src/services/__tests__/scheduler-service.test.ts
stbrody: given the single-thread nature of JS it probably actually is perfectly safe, but for my C++ trained brain that is paranoid about thread safety, could you declare the `runChecks` function before calling `schedulerService.start()`?  Would make the test flow more intuitive as well IMO

config/default.json
```diff
@@ -2,6 +2,7 @@
   "mode": "bundled",
   "port": 8081,
   "anchorControllerEnabled": false,
+  "schedulerIntervalMS": 300000,
   "cronExpression": "0/5 * * * ? *",
```
stbrody, line 6: can we remove all references to `cronExpression` now everywhere in this repo?

stbrody (approved):
pretty much all my comments other than removing `cronExpression` are extreme nitpicks you should feel free to ignore if you disagree with at all.

src/services/__tests__/scheduler-service.test.ts
stephhuynh18: It does feel really weird to use something before its defined. I have changed the tests so it no longer does this haha.

src/services/__tests__/scheduler-service.test.ts
stephhuynh18: I added the comment =)

config/default.json
stephhuynh18: Removed all references

stephhuynh18:
I am holding off on merging this as PLAT needs to do some work to change the config

config/default.json
```diff
@@ -2,7 +2,7 @@
   "mode": "bundled",
   "port": 8081,
   "anchorControllerEnabled": false,
-  "cronExpression": "0/5 * * * ? *",
+  "schedulerIntervalMS": 300000,
```
stbrody, line 5: this isn't the interval when batches run, but just the interval for when the scheduled checks if a batch *should* run, right?  In that case should we make the interval shorter?  Like every minute?  Or even every 10 seconds?

src/models/request-presentation.ts
```diff
@@ -60,7 +58,7 @@ export class RequestPresentation {
           message: request.message,
           createdAt: request.createdAt.getTime(),
           updatedAt: request.updatedAt.getTime(),
-          scheduledAt: awsCronParser.next(cron, new Date()),
+          scheduledAt: new Date(Date.now() + this.schedulerIntervalMS),
```
stbrody, line 61: is this really right?  Isn't the schedulerInterval just how often it checks, not how often anchors will actually happen?  I feel like we might just want to remove `scheduledAt` entirely as there's now no way to predict exactly when an anchor will happen now that we're scheduling it based on usage rather than a fixed interval.

@oed

stbrody: (to be clear, we should file a ticket to remove `scheduledAt`, not do it as part of this PR)

config/default.json
stephhuynh18: Yes I agree. I will be changing this once we are using the scheduler service to emit anchor events. Currently this interval is set to the same amount of time as the cronExpression to preserve current functionality

src/models/request-presentation.ts
ukstv: I concur.

stephhuynh18:
Merging as `scheduler_interval_ms` was applied to dev

PR #662, opened 2022-06-23
feat: emit launch anchor event if ready

stephhuynh18:
* Added Event Producer Service (I used to called it Event Service but that caused problems with tsyringe that i do not understand)
* Added SQS implementation of Event producer Service
* Added createAnchorEventIfReady (not used) 

Should have split this up into a few PRS my bad

### DIFF
--- a/src/services/anchor-service.ts
+++ b/src/services/anchor-service.ts
@@ -29,6 +29,7 @@ import {
   IpfsMerge,
 } from '../merkle/merkle-objects.js'
 import type { Connection } from 'typeorm'
+import { v4 as uuidv4 } from 'uuid'
 
 export const READY_TIMEOUT = 1000 * 60 * 15
 
@@ -223,7 +224,7 @@ export class AnchorService {
   /**
    * Emits and anchor event if
    */
-  public async createAnchorEventIfReady(): Promise<void> {
+  public async emitAnchorEventIfReady(): Promise<void> {
     const readyRequests = await this.requestRepository.findByStatus(RS.READY)
     const readyDeadline = Date.now() - READY_TIMEOUT
 
@@ -246,7 +247,7 @@ export class AnchorService {
       }
     }
 
-    await this.eventProducerService.emitAnchorEvent()
+    await this.eventProducerService.emitAnchorEvent(uuidv4().toString())
 
     return
   }
--- a/src/services/anchor-service.ts
+++ b/src/services/anchor-service.ts
@@ -236,17 +236,14 @@ export class AnchorService {
       await this.requestRepository.updateRequests({ status: RS.READY }, readyRequests)
 
       // TODO: alert we are going to retry
+    } else {
+      const streamLimit = Math.pow(2, this.config.merkleDepthLimit)
 
-      await this.eventProducerService.emitAnchorEvent()
+      const updatedRequests = await this.requestRepository.findAndMarkReady(streamLimit)
 
-      return
-    }
-    const streamLimit = Math.pow(2, this.config.merkleDepthLimit)
-
-    const updatedRequests = await this.requestRepository.findAndMarkReady(streamLimit)
-
-    if (updatedRequests.length === 0) {
-      return
+      if (updatedRequests.length === 0) {
+        return
+      }
     }
 
     await this.eventProducerService.emitAnchorEvent()
--- a/src/services/anchor-service.ts
+++ b/src/services/anchor-service.ts
@@ -16,7 +16,7 @@ import { AnchorRepository } from '../repositories/anchor-repository.js'
 import { RequestRepository } from '../repositories/request-repository.js'
 
 import { IpfsService } from './ipfs-service.js'
-import { EventProducerService } from './event-producer-service.js'
+import { EventProducerService } from './event-producer/event-producer-service.js'
 import { CeramicService } from './ceramic-service.js'
 import { BlockchainService } from './blockchain/blockchain-service.js'
 import { inject, singleton } from 'tsyringe'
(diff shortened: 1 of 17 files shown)


src/services/event-producer/sqs/sqs-event-producer-service.ts
```diff
@@ -0,0 +1,40 @@
...
+import { SQSClient, SendMessageCommand } from '@aws-sdk/client-sqs'
+import { EventProducerService } from '../event-producer-service.js'
+import { v4 as uuidv4 } from 'uuid'
+
+@singleton()
+export class SQSEventProducerService implements EventProducerService {
+  private readonly sqsClient: SQSClient
+  constructor(@inject('config') private config?: Config) {
+    this.sqsClient = new SQSClient({
+      region: process.env.AWS_REGION,
```
ukstv, line 13: @stephhuynh18 AWS library gets it from process.env by its own, no? Are the env variables here for visibility?

src/services/event-producer/sqs/sqs-event-producer-service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Emits an anchor event by sending a message to the configured SQS
+   */
+  public async emitAnchorEvent(): Promise<void> {
+    await this.sqsClient.send(
+      new SendMessageCommand({
+        MessageBody: uuidv4().toString(),
+        MessageGroupId: 'anchor',
+        QueueUrl: process.env.AWS_SQS_URL,
```
ukstv, line 29: Config is not respected. Is it ok?

src/services/anchor-service.ts
```diff
@@ -216,6 +220,40 @@ export class AnchorService {
...
+      const earliestNotTimedOut = readyDeadline < readyRequests[0].updatedAt.getTime()
+      if (earliestNotTimedOut) {
+        return
+      }
+
+      await this.requestRepository.updateRequests({ status: RS.READY }, readyRequests)
+
+      // TODO: alert we are going to retry
+
+      await this.eventProducerService.emitAnchorEvent()
```
ukstv, line 240: Can it be
```
if (readyRequests.length > 0) {
const earliestNotTimedOut = readyDeadline < readyRequests[0].updatedAt.getTime()
       if (earliestNotTimedOut) {
         return
       }

       await this.requestRepository.updateRequests({ status: RS.READY }, readyRequests)

       // TODO: alert we are going to retry
} else {
const streamLimit = Math.pow(2, this.config.merkleDepthLimit)

     const updatedRequests = await this.requestRepository.findAndMarkReady(streamLimit)

     if (updatedRequests.length === 0) {
       return
     }
}
await this.eventProducerService.emitAnchorEvent()
```

src/services/event-producer/sqs/sqs-event-producer-service.ts
stephhuynh18: :o i didn't know that i'll remove this config thanks

src/services/event-producer/sqs/sqs-event-producer-service.ts
stephhuynh18: Could you elaborate pls =)

ukstv: For sure. Sorry for that. You are passing the config into the class, but this SQS queue identifier is still from the env variable. This makes me a little bit nervous, as it splits configuration between env variables and config file, thus making deployment more complicated.

The env variables above are handled by the AWS library AFAIK. This one - I am not sure. _Maybe_ it makes sense to have it in the configuration file.

src/services/event-producer/event-producer-service.ts
```diff
@@ -0,0 +1,15 @@
+import 'reflect-metadata'
+import { Config } from 'node-config-ts'
+import { inject, singleton } from 'tsyringe'
+import { SQSClient, SendMessageCommand } from '@aws-sdk/client-sqs'
```
smrz2001, line 4: Can this import be removed from this file since it is a generic interface?

src/services/event-producer/sqs/sqs-event-producer-service.ts
```diff
@@ -0,0 +1,34 @@
...
+export class SQSEventProducerService implements EventProducerService {
+  private readonly sqsClient: SQSClient
+  constructor(@inject('config') private config?: Config) {
+    this.sqsClient = new SQSClient({})
+  }
+
+  /**
+   * Emits an anchor event by sending a message to the configured SQS
+   */
+  public async emitAnchorEvent(): Promise<void> {
```
smrz2001, line 18: Would it make sense for this method to accept the body as an argument? That way the caller can decide what to send in the event without needing to worry about the underlying queue technology.

Then, you'll also be able to read the event off the queue in your unit tests and make sure the body was correct.

src/services/event-producer/sqs/sqs-event-producer-service.ts
```diff
@@ -0,0 +1,34 @@
...
+    await this.sqsClient.send(
```
smrz2001, line 19: Might be good to add a to-do here for later to switch to localstack for unit tests. Even though the current approach was my recommendation, I don't quite like the idea of a unit test environment that's not fully self-contained, and has to reach out to AWS for something. That just adds a whole host of dependencies for anyone trying to run the unit tests, like credentials, etc.

src/services/event-producer/sqs/sqs-event-producer-service.ts
stephhuynh18: rerequested review to look at config changes

ukstv (approved):
🏇

PR #665, opened 2022-06-27
Fix: flaky ceramic integration tests

stephhuynh18:
The "Consensus for anchors" integration tests periodically fail. It seems that the CAS is not anchoring the latest update. My suspicion (could not reproduce locally) is that the CAS ceramic node isn't receiving the latest update before the anchor. So my fix is to wait for the latest update and then anchor. Timing in tests is a very tricky thing. This may not work but we won't really know until we try.

### DIFF
--- a/src/__tests__/ceramic_integration.test.ts
+++ b/src/__tests__/ceramic_integration.test.ts
@@ -147,28 +147,36 @@ async function anchorUpdate(stream: Stream, anchorService: CeramicAnchorApp): Pr
   await waitForAnchor(stream)
 }
 
-async function waitForAnchor(stream: Stream): Promise<void> {
-  await stream
-    .pipe(
-      filter((state) => [AnchorStatus.ANCHORED, AnchorStatus.FAILED].includes(state.anchorStatus)),
-      take(1)
-    )
-    .toPromise()
-}
-
-async function waitForUpdate(stream: Stream, update: CID, timeoutMS = 30 * 1000): Promise<void> {
+async function waitForAnchor(stream: Stream, timeoutMS = 30 * 1000): Promise<void> {
   await firstValueFrom(
     stream.pipe(
-      filter((state) => {
-        return state.log[state.log.length - 1].cid.toString() === update.toString()
+      timeout({
+        each: timeoutMS,
+        with: () =>
+          throwError(
+            () => new Error(`Timeout waiting for stream ${stream.id.toString()} to become anchored`)
+          ),
       }),
+      filter((state) => [AnchorStatus.ANCHORED, AnchorStatus.FAILED].includes(state.anchorStatus))
+    )
+  )
+}
+
+async function waitForTip(stream: Stream, tip: CID, timeoutMS = 30 * 1000): Promise<void> {
+  await firstValueFrom(
+    stream.pipe(
       timeout({
         each: timeoutMS,
         with: () =>
           throwError(
             () =>
-              new Error(`Timeout waiting for ceramic to receive update cid ${update.toString()}`)
+              new Error(
+                `Timeout waiting for ceramic to receive cid ${tip.toString()} for stream ${stream.id.toString()}`
+              )
           ),
+      }),
+      filter((state) => {
+        return state.log[state.log.length - 1].cid.toString() === tip.toString()
       })
     )
   )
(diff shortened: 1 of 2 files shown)


src/__tests__/ceramic_integration.test.ts
```diff
@@ -387,6 +388,16 @@ describe('Ceramic Integration Test', () => {
         const doc2 = await TileDocument.load(ceramic2, doc1.id)
         await doc2.update(updatedContent, null, { anchor: false })
 
+        // Make sure that the ceramic CAS has received the newest version
+        const casDocRef = await casCeramic1.loadStream(doc1.id)
+        await firstValueFrom(
+          casDocRef.pipe(
+            filter((state) => {
+              return state.next?.content.foo === updatedContent.foo
```
stbrody, line 396: I'm worried about if this condition is never hit the whole test will hang.  Can we add a 30 second timeout to this?  Might want to make a helper function that can be used in both tests and given a condition to check for in the state as well as a timeout

src/__tests__/ceramic_integration.test.ts
```diff
@@ -425,6 +436,16 @@ describe('Ceramic Integration Test', () => {
...
+            })
+          )
+        )
+
         await anchorUpdate(doc1, cas1)
 
         expect(doc1.state.anchorStatus).toEqual(AnchorStatus.ANCHORED)
```
stbrody, line 451: I think the checks on lines 454-456 are subject to the same problem where the pubsub message may just not have made it yet.

stbrody (commented):
I think this is a very good hypothesis as to why the tests are failing sometimes - the test is just moving faster than the pubsub message is delivered.

src/__tests__/ceramic_integration.test.ts
stephhuynh18: yup will do

src/__tests__/ceramic_integration.test.ts
```diff
@@ -155,6 +156,24 @@ async function waitForAnchor(stream: Stream): Promise<void> {
     .toPromise()
 }
 
+async function waitForUpdate(stream: Stream, update: CID, timeoutMS = 30 * 1000): Promise<void> {
+  await firstValueFrom(
+    stream.pipe(
+      filter((state) => {
+        return state.log[state.log.length - 1].cid.toString() === update.toString()
+      }),
+      timeout({
```
stbrody, line 165: does the `timeout` have to be first in the `pipe` to be processed?  I don't know enough about how RXJS works, but I'd have thought the `pipe` processes each entry one-at-a-time and so the `timeout` wouldn't be processed until after the `filter` resolved?  Maybe not though - did you test the timeout behavior?  If you did then it's fine

src/__tests__/ceramic_integration.test.ts
stbrody: ping.  Can we fix that one the same way?

src/__tests__/ceramic_integration.test.ts
stephhuynh18: I can definitely see how you would think that. From the docs `timeout` errors if Observable does not emit a value in given time span. So I think it has to come after the filter. I have tested this works.

src/__tests__/ceramic_integration.test.ts
stephhuynh18: yup sorry got confused

src/__tests__/ceramic_integration.test.ts
```diff
@@ -398,7 +429,7 @@ describe('Ceramic Integration Test', () => {
       60 * 1000 * 2
     )
 
-    test('Anchor discovered through pubsub', async () => {
+    test.only('Anchor discovered through pubsub', async () => {
```
stbrody, line 432:
```suggestion
    test('Anchor discovered through pubsub', async () => {
```

stbrody: left over from testing I assume

stbrody (approved):
lgtm

src/__tests__/ceramic_integration.test.ts
stephhuynh18: good catch!

PR #668, opened 2022-07-04
feat: add scheduler mode and anchor READY requests

stephhuynh18:
* add the scheduler mode
* anchoring READY requests instead of PENDING
* Remove stream limit checks 
* Manually tested scheduler mode

### DIFF
--- a/src/repositories/request-repository.ts
+++ b/src/repositories/request-repository.ts
@@ -160,15 +160,19 @@ export class RequestRepository extends Repository<Request> {
   }
 
   /**
-   * Marks requests as READY if (in the order of precendence):
-   *  1. there are PROCESSING requests that need to be anchored and retried (the maximum anchoring delay has elapsed and the request hasn't been updated in a long time)
-   *  2. there are PENDING requests that need to be anchored (the maximum anchoring delay has elasped)
-   *  3. there are streamLimit streams needing an anchor (prioritizing PROCESSING requests that need to be retried, then PENDING requests)
+   * Marks requests as READY. The following requests may be marked as ready in the order of precendence:
+   *  1. There are FAILED requests that were not failed because of conflict resolution
+   *  2. there are PROCESSING requests that need to be anchored and retried (the maximum anchoring delay has elapsed and the request hasn't been updated in a long time)
+   *  3. there are PENDING requests that need to be anchored (the maximum anchoring delay has elasped)
+   * These requests are only marked as ready if there are streamLimit streams needing an anchor OR the earliest chosen request is about to expire
+   *
    * Returns the original requests that were marked as READY
    */
   public async findAndMarkReady(streamLimit: number): Promise<Request[]> {
     const anchoringDeadline = new Date(Date.now() - MAX_ANCHORING_DELAY_MS)
     const processingDeadline = new Date(Date.now() - PROCESSING_TIMEOUT)
+    const earliestDateToRetry = new Date(Date.now() - FAILURE_RETRY_WINDOW)
+
     const isolationLevel =
       this.connection.options.type === 'sqlite' ? 'SERIALIZABLE' : 'REPEATABLE READ'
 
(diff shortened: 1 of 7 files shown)


src/services/anchor-service.ts
```diff
@@ -479,7 +460,17 @@ export class AnchorService {
 
     if (unprocessedRequests.length > 0) {
       logger.debug(
-        `There were ${unprocessedRequests.length} unprocessed requests that didn't make it into this batch.  Leaving them in PENDING state as they were.`
+        `There were ${unprocessedRequests.length} unprocessed requests that didn't make it into this batch.  Marking them as PENDING.`
```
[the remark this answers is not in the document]
stbrody: I see your TODO comment on the next line now.  Feel free to file a follow-up ticket for this and include the ticket number in the TODO message

src/repositories/request-repository.ts
```diff
@@ -183,16 +192,28 @@ export class RequestRepository extends Repository<Request> {
...
         .getMany()
 
       // Do not anchor if the earliest request isn't expired and there isn't enough streams
       const earliestIsNotExpired =
         streamsToAnchor.length > 0 && streamsToAnchor[0].createdAt >= anchoringDeadline
-      if (earliestIsNotExpired && streamsToAnchor.length < streamLimit) {
+      if (
+        streamsToAnchor.length === 0 ||
+        (earliestIsNotExpired && streamsToAnchor.length < streamLimit)
+      ) {
```
stbrody, line 216: Can we add some logging here to make clear what is happening?

stbrody (commented):
all my comments are pretty small nitpicks, overall it's looking good.

One thing I wasn't sure if I missed - before we have the code to actually schedule anchor batches from AWS, don't we still need the `anchor` mode to do both halves of the process - marking requests as READY and then anchoring READY requests?  Is that already there and I missed it?  Trying to figure out the exact order of operations for synchronizing changes to CAS code with changes to infra configuration and how we actually atomically switch to the new behavior

src/repositories/request-repository.ts
stephhuynh18: The logic seems to be quite confusing here so I'm going to make some changes and add logging

src/services/__tests__/anchor-service.test.ts
```diff
@@ -218,36 +217,6 @@ describe('anchor service', () => {
     expect(anchors[3].path).toEqual('1/1')
   })
 
-  test('Too few anchor requests', async () => {
```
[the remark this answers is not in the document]
stephhuynh18: Originally I removed this test because `requestRepository.findAndMarkReady` determined if there aren't enough requests (and I tested it there), but since we are still in a mode where `anchorRequests` does both halves I have restored this test.

src/services/anchor-service.ts
```diff
@@ -236,9 +216,10 @@ export class AnchorService {
 
       await this.requestRepository.updateRequests({ status: RS.READY }, readyRequests)
```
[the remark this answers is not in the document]
stephhuynh18: Added comment but to repeat here: since the expiration of ready requests are determined by their "updated_at" field, update the requests again to indicate that a new anchor event has been emitted

src/services/anchor-service.ts
```diff
@@ -80,25 +80,27 @@ export class AnchorService {
...
-    // be picked up by the next anchor batch.
-    const streamLimit = Math.pow(2, this.config.merkleDepthLimit)
-    const requestLimit = streamLimit * 100
+    logger.imp('Anchoring ready requests...')
     logger.debug(`Loading requests from the database`)
-    const requests: Request[] = await this.requestRepository.findNextToProcess(requestLimit)
+
+    const readyRequests = await this.requestRepository.findByStatus(RS.READY)
+
+    if (!triggeredByAnchorEvent && readyRequests.length === 0) {
```
stbrody, line 96: This `anchorRequests` function now seems to be doing two fairly different things - anchoring ready requests, but also setting pending requests to ready (when running in the current mode where anchors aren't scheduled separately).  I feel like we should be able to break this up better and make more clear separation of responsibilities and control flow.  I'd love to see this function renamed to `anchorReadyRequests` and always be able to assume that the requests it needs to anchor are already in the READY status.  To do that, we'd need to move this block here up higher in the call stack.  I think higher up you have more natural codepath divergence between whether we're running in "v1" or "v2" mode, where this logic may fit more naturally.

Feel free to disagree though, at this point you're more familiar with the structure of this codebase than I am

stbrody: I'm imagining code that looks something like:

```
if (!triggeredByEvent) {
  markRequestsReadyIfNeeded()
}
anchorReadyRequests()
```

stbrody: maybe actually this function could *be* the higher-level thing that runs the pseudocode I just outlined, but with `_anchorRequests` renamed to `_anchorReadyRequests`, and a helper function added for cleaning up the `if (!triggeredByEvent)` code for making requests ready

stephhuynh18: yup i like this. I'm going to leave `_anchorRequests` as is because its a helper function and is used heavily in the tests. But i've added an `anchorReadyRequests` and `anchorRequests`.  `anchorRequests` retains its original functionality where requests start as PENDING. `anchorReadyRequests` assume the PENDING requests have already moved to READY. When we are cleaning this up later, we just need to remove `anchorRequests` and replace it with `anchorReadyRequests. 

When we trigger an anchor via the server, is that anchoring READY requests or PENDING requests. We can leave that decision to later when we fully transition to CAS V2.

stbrody (approved):
LGTM!

PR #676, opened 2022-07-08
Retrying on uncaught errors

stephhuynh18:
A try catch to return requests to pending explicitly during failure. The requests may already in pending but once we turn on CAS v2 they will be in processing


src/services/anchor-service.ts
```diff
@@ -1,6 +1,6 @@
 import { CID } from 'multiformats/cid'
 
-import { RequestStatus as RS } from '../models/request-status.js'
+import { RequestStatus, RequestStatus as RS } from '../models/request-status.js'
```
stbrody, line 3: can we import `RequestStatus` just once instead of twice under different names?

src/services/anchor-service.ts
```diff
@@ -145,38 +145,53 @@ export class AnchorService {
       }
     }
 
-    logger.imp(`Creating Merkle tree from ${candidates.length} selected streams`)
-    const merkleTree = await this._buildMerkleTree(candidates)
+    try {
```
stbrody, line 148: Can we put the body of the `try` into a `_anchorCandidates()` helpful function to improve readability?

src/services/anchor-service.ts
```diff
@@ -145,38 +145,53 @@ export class AnchorService {
...
-      alreadyAnchoredRequestsCount: groupedRequests.alreadyAnchoredRequests.length,
-      anchoredRequestsCount: numAnchoredRequests,
-      conflictingRequestCount: groupedRequests.conflictingRequests.length,
-      failedRequestsCount: groupedRequests.failedRequests.length,
-      failedToPublishAnchorCommitCount: merkleTree.getLeaves().length - anchors.length,
-      unprocessedRequestCount: groupedRequests.unprocessedRequests.length,
-      candidateCount: candidates.length,
-      anchorCount: anchors.length,
+      logger.imp(`Service successfully anchored ${anchors.length} CIDs.`)
+      return {
```
stbrody, line 171: We should still try to log an anchor batch summary message even on failed batches.  So I think what that probably means is moving this return statement out of the try-catch, but making sure that on the error path all the accepted requests get counted as failed.

src/services/anchor-service.ts
```diff
@@ -145,38 +145,53 @@ export class AnchorService {
...
+      )
+
+      // TODO(NET-1623): Add alert that something went wrong and we are retrying
+
+      await this.requestRepository.updateRequests(
+        { status: RequestStatus.PENDING },
+        groupedRequests.acceptedRequests
+      )
+
+      throw err
```
stbrody, line 194: what happens to the overall anchor batch when this throws?

stephhuynh18: we throw on a lot of cases (ex. ipfs error, insufficient funds, eth provider errors) I'm assuming our current alert system depends on this. The overall anchor batch gets lost forever, nothing persists it. The requests fail in w.e status they are left off in (PROCESSING, FAILED or COMPLETED)

src/services/anchor-service.ts
stephhuynh18: To preserve throwing the errors I instead created a helper function that creates and logs the anchor summary. This helper function is called instead of retuning the anchor summary. 

In regards to "making sure that on the error path all the accepted requests get counted as failed", I have instead added a field called `retryingRequestsCount` for these. I am hesitant to add them to the failed count because `failed` currently means that we have purposeful marked the request as failed. `retryingRequestsCount` means something went wrong. I do not feel strongly tho so if you prefer for me to add them to the failed count I will definitely do so.

src/services/anchor-service.ts
stephhuynh18: yes will definitely do so

src/services/anchor-service.ts
```diff
@@ -139,19 +162,52 @@ export class AnchorService {
...
+
+      // groupedRequests.failedRequests include all requests that were failed when loading the streams.
+      // This includes failed requests associated to candidates. Since requests may have been marked failed
+      // during the anchoring of the candidates, there may be more failed requests associated with candidates than
+      // before. This is why we must do the following calculation as to not double count some failed requests.
+      const candidateFailedRequestsAfterAnchoring = candidates
+        .map((candidate) => candidate.failedRequests)
+        .flat().length
+
+      const results: Partial<AnchorSummary> = {
```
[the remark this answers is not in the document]
stephhuynh18: yes i agree. I'm going to try to simplify it by marking everything as FAILED. My intention was to preserve what statuses mean currently but that translated into this complexity.

src/services/anchor-service.ts
```diff
@@ -139,19 +160,33 @@ export class AnchorService {
...
     }
 
+    try {
+      const results = await this._anchorCandidates(candidates)
+      logAnchorSummary(groupedRequests, candidates, results)
+      return
+    } catch (err) {
+      // TODO(NET-1623): Add alert that something went wrong and we are retrying
+      logger.warn(
+        `Updating PROCESSING requests to FAILED so they are retried in the next batch because an error occured while creating the anchors: ${err}`
```
stbrody, line 174: Why FAILED and not PENDING if the plan is to retry them?

stephhuynh18: Well at this point PENDING + FAILED both get retried immediately and I do not have any strong preference for either. I can see how PENDING might make more sense when trying to understand everything so I'll change it to that.

stbrody: Yeah I thought the plan was to eventually stop retrying FAILED requests and only mark requests as FAILED when we know they have truly failed and won't recover on a retry.  Then for things we want to retry we move them back to PENDING.  Did you have a different end state in mind?

stephhuynh18: Ah wait after revising it I see why I changed them as FAILED. Because I include them in the `failedRequestCount` in the logged `anchorSummary`. I'll still include them in the `failedRequestCount` even if they weren't marked FAILED cause technically they did fail.

stephhuynh18: > Then for things we want to retry we move them back to PENDING. Did you have a different end state in mind?

We will probably introduce so more complicated retry logic later when things get simplified and separated. Probably a RETRY state or something like that. But thats for later.

stbrody: Should we split out `failedRequestCount` into two separate counters?  One for permanently failed requests and one for requests that failed but will be retried?  We can add or remove entries from the anchorSummary log message as we want.

stephhuynh18: Current:
`failedRequestCount` = purposefully marked failed requests + all requests that are being reverted to PENDING

Your proposal: 
`failedRequestCount` = purposefully marked failed requests 
`permanentlyFailedRequest` = `conflictingRequestCount` (already tracked)
`failedButRetriableCount` = `failedRequestCount` - `conflictingRequestCount` + all requests that are being reverted to PENDING (note that retriable does not mean the request will be retried. If the request expired that it will not)

Based on what is in the anchor summary in the current situation we can infer the amount of `failedButRetriableCount` through `failedRequestCount - conflictingRequestCount`. 

Questions:
- The additional information your proposal gives is the amount of requests that we have purposefully marked as failed during our processes. Is this information what we want? 
- Do we also want the number of requests that are being reverted to PENDING? 
- Are we interested in have a specific label for `retriableRequestsCount` even if it can be inferred for clarity? And if so is it implicit knowledge that some of these requests won't be retried because they have expired in the next batch

Feel free to ping me or hop in a chat if this is confusing.

stbrody: I didn't realize we could infer the difference from what we already have.  I don't feel strongly about what we do here - I think we should do whatever is simplest, and we shouldn't be afraid to redefine some of these fields in the log summary if it helps with long-term clarity and simplicity.

stephhuynh18: Based on this exchange i think i'll add a label for `retriableRequestsCount`. I can add a comment explaining this there as well for clarity so we don't forget

stbrody (approved):
I still really wish we could simplify that catch block, but this is probably good enough for now

ukstv (approved):

PR #677, opened 2022-07-08
feat: Add transaction mutex and acquire before creating and sending tx

stephhuynh18:
* Using postgres advisory lock implement a transaction based mutex
* Call before creating and sending transaction
* Add tests


src/repositories/request-repository.ts
```diff
@@ -184,80 +197,92 @@ export class RequestRepository extends Repository<Request> {
...
-        // TODO(NET-1623): Add alert here that we marked expired and processing requests as READY
-        return requests
-      }
+        logger.debug(
+          'Not updating any requests to READY because there are not enough streams for a batch and the earliest request is not expired'
+        )
+        return []
+      })
+      .catch((err) => {
+        if (err?.code === REPEATED_READ_SERIALIZATION_ERROR) {
```
stephhuynh18, line 279: Note this is a lot of formatting change because of this catch

src/repositories/request-repository.ts
```diff
@@ -18,6 +19,9 @@ const ANCHOR_DATA_RETENTION_WINDOW = 1000 * 60 * 60 * 24 * 30 // 30 days
 export const MAX_ANCHORING_DELAY_MS = 1000 * 60 * 60 * 12 //12H
 export const PROCESSING_TIMEOUT = 1000 * 60 * 60 * 3 //3H
 export const FAILURE_RETRY_WINDOW = 1000 * 60 * 60 * 48 // 48H
+const TRANSACTION_MUTEX_ID = 4532
```
stbrody, line 22: any reason for this value?  Totally arbitrary?

src/repositories/request-repository.ts
```diff
@@ -97,27 +101,36 @@ export class RequestRepository extends Repository<Request> {
...
+            `A problem occured when updated requests to PROCESSING. Only ${results.affected}/${requests.length} requests were updated`
+          )
+        }
+
+        return requests
+      })
+      .catch((err) => {
+        if (err?.code === REPEATED_READ_SERIALIZATION_ERROR) {
+          // TODO (NET-1623): Add alert here that we have to retry the find and mark processing (very odd case)
+          return this.findAndMarkAsProcessing()
```
stbrody, line 129: maybe add a tiny sleep to avoid hammering the database?  100ms?

src/repositories/request-repository.ts
```diff
@@ -184,80 +197,92 @@ export class RequestRepository extends Repository<Request> {
...
-      }
+        logger.debug(
+          'Not updating any requests to READY because there are not enough streams for a batch and the earliest request is not expired'
+        )
+        return []
+      })
+      .catch((err) => {
+        if (err?.code === REPEATED_READ_SERIALIZATION_ERROR) {
+          // TODO (NET-1623): Add alert here that we have to retry the find and mark ready again (very odd case)
+          return this.findAndMarkReady(maxStreamLimit, minStreamLimit)
```
stbrody, line 281: ditto re:short sleep

src/services/anchor-service.ts
```diff
@@ -156,10 +156,10 @@ export class AnchorService {
     const merkleTree = await this._buildMerkleTree(candidates)
 
     // create and send ETH transaction
-    logger.debug('Preparing to send transaction to put merkle root on blockchain')
-    const tx: Transaction = await this.blockchainService.sendTransaction(
-      merkleTree.getRoot().data.cid
-    )
+    const tx: Transaction = await this.requestRepository.withTransactionMutex(() => {
```
stbrody, line 159: what will the behavior wind up being if we keep failing to acquire the mutex forever?

stbrody (approved):

src/repositories/request-repository.ts
stephhuynh18: yes totally arbitrary haha

src/services/anchor-service.ts
stephhuynh18: As is, this will keep retrying indefinitely to acquire the mutex.

This scenario should not caused be cause by `this.blockchainService.sendTransaction` because this should timeout after 3 attempts. 

But it can be caused by other CAS's always acquiring the mutex. I hope this doesn't happen. This can be helped via some queueing system/nonce management which we can think about for another PR. 

I have built in max attempts and delays which we could configure. I also could add a timeout if needed. But that is uncertain right now. In this scenario the batch will fail. Do you have any thoughts about this or I can bring this up in a discussion at a later time. 

What I can do in the meantime is add alerts every time a CAS cannot acquire a mutex after >5 (arbitrary number) attempts. That way we can know how big of a problem this is.

src/repositories/request-repository.ts
stbrody: Can we add a comment or make it like `1234` so it's more clearly arbitrary?

src/services/anchor-service.ts
stbrody: yeah, could also happen if somehow the mutex gets stuck in a locked state.

I see that this process will keep retrying forever, I was wondering more from the perspective of the overall flow.  Is there some higher-level timeout in AWS that will fire eventually?  Or the process keep running forever?  Will we notice if it keeps running forever?  Will this trigger an alert?  What happens when the scheduler service fires off another anchor worker?

stephhuynh18: The mutex should not get stuck locked forever unless something is wrong with postgres or typeorm (more likely). The mutex gets autoreleased at the end of the DB transaction. Also if a session get dropped all of it's locks are released as well. 

I will ask PLAT about these questions. If another anchor worker is launched while one is waiting for the mutex forever, it too will also wait for the lock forever until someone notices which isn't great

PR #688, opened 2022-07-22
CI/CD integration fixes for the build contacts, and build cas steps

Jovonni:
This fixes the broken steps of the pipeline for building contracts, and cas

## Description
Small PR to fix issues in our pipeline for building the contracts. The contract building was failing because we were installing the deps as a the ```circleci``` user, but then trying to invoke the NPM script as ```root```. Since we were setting the PATH with $HOME, using ```sudo``` changes that ```$HOME``` directory to like ```/home/root``` etc

```yaml
- run:
   name: install dependencies
   command: |
     curl -L https://foundry.paradigm.xyz | bash ## installing as circleci ##
      echo 'export PATH="$HOME/.foundry/bin:$PATH"' >> /home/circleci/.bashrc
     source /home/circleci/.bashrc
     echo 'export PATH="$HOME/.foundry/bin:$PATH"' >> $BASH_ENV
     source $BASH_ENV
     foundryup
     forge --version
- run:
     name: build contracts
     command: |
     echo $PATH
     sudo npm run installContractDeps ## invoking as root ##
     sudo npm run buildContract  ## invoking as root ##
```

Also, inside the package.json, the preBuild stage is running before the build stage, and because the repo isn't built yet by this step in the runner's environment, the ```"clean": "rm -rf ./build; rm -rf coverage; rm -rf .nyc_output",``` commands were failing since those artifacts didn't exist yet. 

We can use ```--ignore-scripts``` for the ```build cas``` step, since that won't change the actual package script command. We may not need to run ```clean``` on a fresh build in the pipeline, especially if no build artifacts are being pushed, and since they are .gitignored here:

https://github.com/ceramicnetwork/ceramic-anchor-service/blob/develop/.gitignore

## How Has This Been Tested?

- [x] Local machine
- [x] CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,8 +54,8 @@ jobs:
           command: |
             echo $PATH
             ls -lah $HOME/.foundry/bin 
-            sudo npm run installContractDeps
-            sudo npm run buildContract
+            npm run installContractDeps
+            npm run buildContract
 
       - run:
           name: build cas
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,10 @@ jobs:
 
       - run:
           name: build cas
-          command: sudo npm run build
+          command: |
+            node -v
+            ls -lah ./
+            sudo npm run build
 
       - run:
           name: test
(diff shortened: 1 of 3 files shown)


stephhuynh18 (approved):

stephhuynh18:
Going to merge into my branch and fix tests there

PR #696, opened 2022-08-03
chore: remove batching from anchor (DO NOT MERGE, NET-1711) 

stephhuynh18:
Once scheduler service is up an anchor does not have to batch so will merge this PR once the scheduler service is up and running


v-stickykeys:
To answer your question from standup. I think we can do this once CASv2 is on mainnet. (soon:tm:)

stbrody:
@stephhuynh18 what's the status of this PR?

PR #702, opened 2022-08-09
fix: version should be 0 if not using smart contract for anchor

stephhuynh18:


oed (commented):
Can we simply not include the `version` property if version is 0?

stbrody (approved):

src/services/anchor-service.ts
```diff
@@ -330,14 +330,16 @@ export class AnchorService {
...
+    const ipfsAnchorProof = Object.assign(
+      {
+        blockNumber: tx.blockNumber,
+        blockTimestamp: tx.blockTimestamp,
+        root: merkleRootCid,
+        chainId: tx.chain,
+        txHash: txHashCid,
+      },
+      this.config.useSmartContractAnchors && { version: 1 }
+    )
```
oed, line 342:
```suggestion
    const ipfsAnchorProof = {
      blockNumber: tx.blockNumber,
      blockTimestamp: tx.blockTimestamp,
      root: merkleRootCid,
      chainId: tx.chain,
      txHash: txHashCid,
    }
    if (this.config.useSmartContractAnchors) ipfsAnchorProof.version = 1
```
Can we make this a bit easier to read please?

oed (approved):
lgtm, one suggestion

src/services/anchor-service.ts
oed: Oh, typescript might not like that?

stephhuynh18: only because the packages need to be updated to the latest version. I can work around

PR #703, opened 2022-08-11
feat: use bytes32 in smart contract

stephhuynh18:
I discard the first 4 bytes of the CID [version, code, hash code, digest size]. This is 2 more bytes than described in the ticket. Please let me know if this is okay.

### DIFF
--- a/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
+++ b/src/services/blockchain/ethereum/ethereum-blockchain-service.ts
@@ -21,7 +21,7 @@ const NUM_BLOCKS_TO_WAIT = 4
 export const MAX_RETRIES = 3
 
 const POLLING_INTERVAL = 15 * 1000 // every 15 seconds
-const ABI = ['function anchor(bytes)']
+const ABI = ['function anchorDagCbor(bytes32)']
 
 class WrongChainIdError extends Error {
   constructor(expected: number, actual: number) {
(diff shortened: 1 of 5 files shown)


src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -278,21 +278,24 @@ export class EthereumBlockchainService implements BlockchainService {
...
-    const rootStrHex = rootCid.toString(base16)
-    const hexEncoded = '0x' + (rootStrHex.length % 2 == 0 ? rootStrHex : '0' + rootStrHex)
-    logger.imp(`Hex encoded root CID ${hexEncoded}`)
-
     logger.debug('Preparing ethereum transaction')
     const baseNonce = await this.wallet.provider.getTransactionCount(this.wallet.address)
 
     if (!this.config.useSmartContractAnchors) {
+      const rootStrHex = rootCid.toString(base16)
+      const hexEncoded = '0x' + (rootStrHex.length % 2 == 0 ? rootStrHex : '0' + rootStrHex)
```
oed, line 286:
```suggestion
      const hexEncoded = '0x' + uint8arrays.toString(rootCid.bytes.slice(4), 16)
```
I would do something like this to be sure

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -278,21 +278,24 @@ export class EthereumBlockchainService implements BlockchainService {
...
       return {
         to: this.wallet.address,
         data: hexEncoded,
         nonce: baseNonce,
       }
     }
-    const transactionRequest = await this._contract.populateTransaction.anchor(hexEncoded)
+
+    const transactionRequest = await this._contract.populateTransaction.anchorDagCbor(
+      rootCid.bytes.slice(4)
```
oed, line 297: Does this function accept a uint8array? If so nice, otherwise I guess it needs a hex encoded string?

oed (commented):
> I discard the first 4 bytes of the CID [version, code, hash code, digest size]. This is 2 more bytes than described in the ticket. Please let me know if this is okay.

Yep this is correct.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
stephhuynh18: Nope doesn't seem to. I'll use the hex encoded string

oed (approved):
lgtm

PR #721, opened 2022-08-24
fix(runner): log ok to start app

v-stickykeys:
## Description

Fixes a bug where the app doesn't start because `npm run start` doesn't get called.
This PR adds back the console log which is checked before running the npm script.

## How Has This Been Tested?

Tested locally with docker, was able to repro and see the fix work.


smrz2001 (approved):
:shipit:

PR #723, opened 2022-08-24
fix: duplicate key error

stephhuynh18:
We have been using TypeORM save which updated if the entity exists and inserts if it does not. Upsert does this as well but also handles conflicts. 

This should resolve the situation where someone tries to create two of the same requests at once.


ukstv (approved):
🏇

PR #724, opened 2022-08-25
Fix Docker Compose

v-stickykeys:
## Description

This PR fixes docker compose so that CAS properly connects to the IPFS container that gets spun up.
This also fixes our CI which tries to use docker compose to do image verification.

- When ipfs-http-client is used to subscribe to pubsub, it makes an http request to IPFS but expects the uri to start with http or https.
- Also the port used by ceramicnetwork/ipfs-daemon is 5011, not 5001.

## How Has This Been Tested?

Tested locally and was able to repro the issue and see the fix work.

Broken
image

Working
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,9 @@ Ceramic anchor service is a PoC implementation of an anchor service according to
 
 This implementation currently uses the Ethereum blockchain but is built in order to be blockchain agnostic. It is fairly easy to add more modules to support other blockchains as well.  
 
-## Usage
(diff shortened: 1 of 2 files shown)


smrz2001:
Oh huh, was this not working before? Did something change recently? The Docker Compose workflow has been getting used for verification for a little bit now, maybe CAS wasn't connecting to IPFS at that time and so that's why the workflows were passing?

smrz2001 (approved):
🚀

PR #726, opened 2022-08-27
feat(ci): add separate ecr repo for runner image

v-stickykeys:
## Description

This PR pushes the base `Dockerfile` image to ECR in addition to dockerhub, and pushes the `Dockerfile.runner` image to a new ECR repo.

This helps 3BL ensure ECS tasks will not hit rate limits when trying to pull their container image, because it will now come from ECR instead of Dockerhub.

## How Has This Been Tested?

https://github.com/ceramicnetwork/ceramic-anchor-service/runs/8047076547?check_suite_focus=true


smrz2001:
I'm hoping to use Dagger here as well but with two separate images, one being derived from the other, that's a little complicated, which is why I didn't do it before. That might just be a lack of understanding on my part so I might experiment with it some.

smrz2001 (approved):
:shipit:

PR #733, opened 2022-08-31
feat: replace sqs with http for anchor launcher

v-stickykeys:
The cd-manager responsible for anchor workers has an API that we can hit from the CAS scheduler mode. (We are no longer using and SQS implementation to handle this).

This PR replaces the sqs event service with an http one that uses fetchJson.

Note:
- [x] Dev infra is updated to support new env var name which is more agnostic to the event service implementation in use.


v-stickykeys:
Update tests now

src/__tests__/ceramic_integration.test.ts
```diff
@@ -113,6 +115,29 @@ async function makeGanache(startTime: Date, port: number): Promise<Ganache.Serve
...
+    const app = express()
+    app.all('/', (req, res) => {
+      res.send({status: 'success'})
+    })
+    this.server = app.listen(port, () => {
+      console.log(`Listening on port ${port}`)
+    })
+  }
+  stop() {
+    this.server.close()
```
stephhuynh18, line 131: small nitpick: close takes a callback so it returns before everything is nicely shut down. Not waiting shouldn't be too much of a big deal in tests  but to ensure tests run smoothly you can replace this with: 
```
  async stop() {
    return new Promise((resolve) => this.server.close(resolve))
  }
  ```
  
  And then await on line 392

stephhuynh18 (approved):
A few small updates but other than that LGTM

PR #746, opened 2022-09-12
fix: single leaf merkle trees store and link metadata

stephhuynh18:
If there is only one stream, the merkle tree root will have a left node as the stream, a right nodes as the SAME stream, and then a link to the metadata. 

I am not entirely sure if this is okay but all the tests pass so seems fine 🤔.  I could also do the left node as the stream, and the right node as NULL, and a link to the metadata but that requires an interface change and not sure what the implications of that are.


src/__tests__/ceramic_integration.test.ts
```diff
@@ -162,6 +162,7 @@ async function makeCAS(
   configCopy.ceramic.apiUrl = 'http://localhost:' + minConfig.ceramicPort
   configCopy.blockchain.connectors.ethereum.network = 'ganache'
   configCopy.blockchain.connectors.ethereum.rpc.port = minConfig.ganachePort + ''
+  configCopy.useSmartContractAnchors = true
```
stbrody, line 165: I think we still want it false by default right?  Ideally we'd actually run all these tests twice, once with it on and once with it off.  But if we're only running them once for now we should run the version that is most similar to what we run in prod

src/merkle/merkle-tree.ts
```diff
@@ -83,6 +83,13 @@ export class MerkleTree<N, L extends N, M> {
       )
     }
 
+    // if there is only one leaf for the whole tree
+    if (elements.length === 1 && treeDepth === 0) {
+      const merged = await this.mergeFn.merge(elements[0], elements[0], treeMetadata)
```
stbrody, line 88: I don't think this can work, as the anchor commit includes the path through the merkle tree to the leaf node, so only one of the two paths would ever validate successfully.  I'd suggest using `null` for the right child here.

stbrody (changes requested):
Please also update the commit message to clarify that this is behavior change (so `feat` not `chore`) and have the message focus on the data structure changes as the primary change, not the package upgrades

src/__tests__/ceramic_integration.test.ts
stephhuynh18: Yes I agree. Smart contract anchors are required in js-ceramic for ganache test networks with chainid 1337. This requirement is turned on in js-ceramic through a block-threshold that isn't exposed. While I can mock the function so we can skip this verification, doing so turns off all other verifications that we probably want in our tests. So i'll need to make another PR to js-ceramic to add a new chainId to specifically tests non-smart contract anchors.

src/merkle/merkle-tree.ts
stephhuynh18: no problem!

src/__tests__/ceramic_integration.test.ts
stbrody: yeah that sounds like a problem in js-ceramic.  The smart contract anchoring project is still in progress, it should be disabled by default and we should have a flag to turn it on in our tests until we're ready to turn it on by default.

src/merkle/merkle-objects.ts
```diff
@@ -202,10 +202,13 @@ export class IpfsMerge implements MergeFunction<CIDHolder, TreeMetadata> {
...
   async merge(
     left: Node<CIDHolder>,
-    right: Node<CIDHolder>,
+    right: Node<CIDHolder> | null,
     metadata: TreeMetadata | null
   ): Promise<Node<CIDHolder>> {
-    const merged = [left.data.cid, right.data.cid]
+    const merged = [left.data.cid]
+
+    merged.push(right === null ? null : right.data.cid)
```
stbrody, line 210: how did this work if there were an odd number of requests in a batch?  Wouldn't there already be cases where a node had one real child and one null child?

src/merkle/merkle-objects.ts
```diff
@@ -202,10 +202,13 @@ export class IpfsMerge implements MergeFunction<CIDHolder, TreeMetadata> {
 
   async merge(
     left: Node<CIDHolder>,
-    right: Node<CIDHolder>,
+    right: Node<CIDHolder> | null,
     metadata: TreeMetadata | null
   ): Promise<Node<CIDHolder>> {
-    const merged = [left.data.cid, right.data.cid]
+    const merged = [left.data.cid]
```
stbrody, line 208:
```suggestion
    const merged = [left?.data?.cid, right?.data?.cid]
```

stbrody: I think this simplifies things and lets you remove the extra `merged.push` line

stbrody (approved):
lgtm mod question around how unbalanced trees ever worked.

src/merkle/merkle-objects.ts
stephhuynh18: So if there were three requests 

left + middle merge creating A
right would not be touched
A + right + metadata would be merged into root

src/merkle/merkle-objects.ts
stephhuynh18: nice

stbrody: oh wait nevermind, this will add result in `undefined` not `null`

src/merkle/merkle-objects.ts
stbrody: ah I see, whenever there's an imbalance we roll the child up and have it replace the parent

src/merkle/merkle-objects.ts
```diff
@@ -202,10 +202,11 @@ export class IpfsMerge implements MergeFunction<CIDHolder, TreeMetadata> {
 
   async merge(
     left: Node<CIDHolder>,
-    right: Node<CIDHolder>,
+    right: Node<CIDHolder> | null,
     metadata: TreeMetadata | null
   ): Promise<Node<CIDHolder>> {
-    const merged = [left.data.cid, right.data.cid]
+    const merged = [left.data.cid, right?.data?.cid]
```
stbrody, line 208:
```suggestion
    const merged = [left.data.cid, right?.data?.cid || null]
```
Sorry, the version I suggested would wind up putting `undefined` into the tree instead of null, which I don't think can even be expressed in IPLD

PR #748, opened 2022-09-13
feat: max anchoring delay is configurable

stephhuynh18:
Need to also add the env var to our various places


config/default.json
```diff
@@ -8,6 +8,7 @@
   "merkleDepthLimit": 0,
   "minStreamCount": 1,
   "anchorLauncherUrl": "http://localhost:8001",
+  "maxAnchoringDelayMS": 43200000,
```
stbrody, line 11: @oed how would you feel about making the max anchoring window be 6 hours instead of 12.  It will still save us a lot of work/money compared to what we're doing now, but give us a bit more wiggle room to make sure that the requests get anchored within the 24 hour CACAO window.

stbrody: also, do we do anything to prevent applications from setting the CACAO expiration to *less than* 24 hours?  If not that could break a lot of the guarantees we're aiming for

stbrody (approved):

stbrody:
@stephhuynh18 did we ever figure out why this **isn't** causing problems on QA right now?

stephhuynh18:
> @stephhuynh18 did we ever figure out why this **isn't** causing problems on QA right now?

@stbrody When I look at the MIN_STREAM_COUNT it is set to 1 in the scheduler. Meaning as long as there is at least one stream we will anchor. I'm not sure if that was placed recently as a work around or if it was always like that.

stephhuynh18:
max_anchoring_delay_ms env was added to infra. Will merge now

PR #753, opened 2022-09-15
fix: does not crash if even producer is not successful 

stephhuynh18:


v-stickykeys (approved):
Great catch!

smrz2001 (approved):
:shipit:

PR #754, opened 2022-09-16
feat: replace typeorm with knex 

stephhuynh18:
Such a long PR T_T 
I was trying to figure out how to break it up but it wasn't working out. If i were to try i would have to put in a lot of effort to set things up so our tests could run using the different libs and tables/databases which would get thrown away. 

so here we are 

I had to write the migration so that it would create new tables for tests, but also transform the existing tables to a compatible format. Thats why it's so long and weird. 
Some ENV vars have to change because they were typeorm specific 
And since this was such a large PR I took the opportunity to move some things around. 
Subscriber was removed because I don't think it was used (I checked the logs). Instead I added logs for inserting into the db (similar to the update db log)

Please take your time and feel free to tell me to make large changes (I left it as draft because of this). If this is way too large, I'll give it a think and try to figure out how to parse this up)


.env.prod.example
```diff
@@ -13,12 +13,8 @@ ETH_RPC_HOST=http://localhost
...
-TYPEORM_HOST = localhost
-TYPEORM_USERNAME = root
-TYPEORM_PASSWORD = root
-TYPEORM_DATABASE = anchor_db
-TYPEORM_PORT = 5432
-TYPEORM_SYNCHRONIZE = true
-TYPEORM_LOGGING = true
-TYPEORM_ENTITIES = build/models/**/*.js
+DB_HOST = localhost
+DB_USERNAME = root
```
willex, line 17: do you need to store `USERNAME` & `PASSWORD` in the example file? You could just parse from env connection string right?

migrations/20220818164120_create_anchor_and_request.js
```diff
@@ -0,0 +1,102 @@
...
+const createRequestTable = async (knex, tableName) => {
+  await knex.schema.createTable(tableName, (table) => {
+    table.uuid('id').notNullable().unique().primary().defaultTo(knex.raw('uuid_generate_v4()'))
+    table.integer('status').notNullable()
+    table.boolean('pinned').notNullable().defaultTo(false)
+    table.timestamp('created_at', { useTz: false }).notNullable().defaultTo(knex.raw('now()'))
+    table.timestamp('updated_at', { useTz: false }).notNullable().defaultTo(knex.raw('now()'))
+  })
+
+  // knex defaults to varchar(255).
```
stbrody, line 31: I think this is what's happening in the ComposeDB database tables currently.  @willex can you follow up on this?

migrations/20220818164120_create_anchor_and_request.js
```diff
@@ -0,0 +1,102 @@
...
+
+  if (!tableExists) {
+    await createRequestTable(knex, REQUEST_TABLE_NAME)
+  } else {
+    await createRequestTable(knex, TEMP_TABLE_NAME)
+    await verifyTableStructure(knex, TEMP_TABLE_NAME, REQUEST_TABLE_NAME)
+    await knex.schema.dropTable(TEMP_TABLE_NAME)
+  }
+
+  // explicitly rename the doc_id column to stream_id
```
stbrody, line 51: won't this be a problem if the table already exists and has already renamed its column to stream_id?  Wouldn't it actually fail `verifyTableStructure` above?

migrations/20220818164120_create_anchor_and_request.js
```diff
@@ -0,0 +1,102 @@
...
+
+const createAnchorTable = async (knex, tableName) => {
+  await knex.schema.createTable(tableName, (table) => {
+    table.uuid('id').notNullable().unique().primary().defaultTo(knex.raw('uuid_generate_v4()'))
+    table.timestamp('created_at', { useTz: false }).notNullable().defaultTo(knex.raw('now()'))
+    table.timestamp('updated_at', { useTz: false }).notNullable().defaultTo(knex.raw('now()'))
+  })
+
+  await knex.schema
+    // we create this column manually because it was misnamed when we originally used typeorm
```
stbrody, line 65: sorry, what's misnamed about it?

stbrody: oh I see, the trailing space.  Can you add that to the comment please?

migrations/20220818164120_create_anchor_and_request.js
stbrody: I don't know how knex migrations really work, but it seems like they should be able to have some conditional logic to only run if they haven't already run.  Can we pull out the changes to the table structures over what was there previously into their own migration - so first we create the tables as they've always been, then we do a second migration to update them into the new schemas, and make sure it all works when run against all 3 possible prior states (clean with no tables, has existing tables in the old format, has existing tables in the new format)

src/service-metrics/index.ts
```diff
@@ -3,7 +3,7 @@
```
stbrody, line 3: any reason for renaming this file?

src/models/request.ts
```diff
@@ -1,39 +1,33 @@
-import { RequestStatus } from './request-status.js'
-import TypeORM from 'typeorm'
-const { Entity, PrimaryGeneratedColumn, Column, CreateDateColumn, UpdateDateColumn, Unique } =
-  TypeORM
+export const TABLE_NAME = 'request'
+
+// How long we should retain requests that were completed or failed
+export const ANCHOR_DATA_RETENTION_WINDOW = 1000 * 60 * 60 * 24 * 30 // 30 days
```
stbrody, line 4: This seems like a weird place for these constants?  They aren't really related to the Request structure, but are really about the behavior of the request repository and anchor service

src/db-connection.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  if (typeof value === 'string') {
+    return camelCase(value)
+  }
+
+  return value
+}
+
+async function runMigrations(connection: Knex) {
+  const [_, pendingMigrations] = await connection.migrate.list()
```
stbrody, line 27: do you know how knex decides which migrations have already been run or not?

src/db-connection.ts
```diff
@@ -0,0 +1,77 @@
...
+    logger.imp(`Running ${pendingMigrations.length} migrations`)
+    await connection.migrate.latest()
+  }
+}
+
+export async function createDbConnection() {
+  const knexConfig: Knex.Config = {
+    ...config.db,
+    wrapIdentifier: (value, origWrap): string => origWrap(snakeCase(value)),
+    postProcessResponse: (result) => toCamelCase(result),
```
stbrody, line 38: Can you add comments explaining what this is doing and why it's needed?

src/repositories/__tests__/request-repository.test.ts
```diff
@@ -194,6 +126,131 @@ describe('request repository test', () => {
...
+    expect(loadedRequests[0].createdAt.getTime()).toBeLessThan(
+      loadedRequests[1].createdAt.getTime()
+    )
+    expect(loadedRequests[0].cid).toEqual(requests[0].cid)
+    expect(loadedRequests[1].cid).toEqual(requests[1].cid)
+  })
+
+  describe('findRequestsToGarbageCollect', () => {
+    test('Finds requests older than a month', async () => {
+      const threeMonthsAgo = new Date(Date.now() - MS_IN_MONTH * 3)
```
stbrody, line 156: I feel like these garbage collection tests are now much longer and harder to identify the core parts that matter.  Could you put back a helper function (or functions) for creating the requests to make it easier to parse what's happening here to a reader?

Considering the code block from berfore:
```
      generateCompletedRequest(false, false),
      generateCompletedRequest(true, false),
      generateCompletedRequest(false, true),
      generateCompletedRequest(true, true),
```

The nice thing about it was that it's pretty clear that there are 2 possible things we are changing about the requests and we are exhausting the 4 possible combinations of those 2 options.
The downside is that it's not immediately obvious to the reader what the 2 boolean arguments represent, but that's pretty easy to tell by looking at the definition of `generateCompletedRequest`.  Could also consider making the argument an object so the options are named at the call site and it's clearer to readers of just the calling code. E.G.: `generateCompletedRequest({expired: false, failed: false})`

src/repositories/request-repository.ts
```diff
@@ -37,327 +81,256 @@ const countRetryMetrics = (requests: Request[], anchoringDeadline: Date): void =
...
+    const { connection = this.connection } = options
+    const [{ id }] = await connection
+      .table(TABLE_NAME)
+      .insert(request, ['id'])
+      .onConflict('cid')
+      .merge()
+
+    const created = await connection.table(TABLE_NAME).first().where({ id })
+
+    logEvent.db({
```
stbrody, line 103: I can't find where this happened before.  Is this new?

src/repositories/request-repository.ts
```diff
@@ -37,327 +81,256 @@ const countRetryMetrics = (requests: Request[], anchoringDeadline: Date): void =
...
+      action: 'upsert',
+      ...request,
+      createdAt: created.createdAt.getTime(),
+      updatedAt: created.updatedAt.getTime(),
+    })
+
+    return created
+  }
+
+  /**
```
stbrody, line 114: sorry, this file is really hard to review with so much code motion.  Can you at least put the functions back into the same relative order to each other so it's easier to compare functions before and after?

stbrody (commented):
For future reference, I think when you have a big PR that touches a lot of files, that's when it's *even more important* than usual to keep the PR as narrow and focused as possible on just the core change, and really fight the urge to slip in one other little thing while you're in there.  It makes it really hard to review, and also means that if anything breaks after this PR goes in it will be hard to tell if it's due to the TypeOrm->Knex migration or due to any of the other code motion or refactorings that are bundled in here.

migrations/20220818164120_create_anchor_and_request.js
stephhuynh18: yup good call. This will make it a lot cleaner

src/service-metrics/index.ts
stephhuynh18: I'll change this back. As you said in the larger comment I got caught up in the refactoring and ended up slipping things in I shouldn't have. Sorry about that and thanks for bearing with me!

src/models/request.ts
stephhuynh18: I debated about this. The problem I was trying to solve was that these constants are littered throughout the code and are kind of hard to find unless you know the code well. Ex. READY_TIMEOUT was in anchor-service.ts while PROCESSING_TIMEOUT was in request-repo. 

So while I agree this may not be the right place for these constants I would like to make sure we keep them in a consistent place thats easy to find. Maybe I should put them all in settings.ts?

src/db-connection.ts
stephhuynh18: knex has a migration table that keeps track of this

src/db-connection.ts
stephhuynh18: sure but just to explain here this is used to convert identifiers between camel case and snake case instead of having out code written with bot. Its only a preference to keep things consistent within the code but if you and anyone else finds it unnecessary I can remove it.

src/repositories/__tests__/request-repository.test.ts
stephhuynh18: yup will do!

src/repositories/request-repository.ts
stephhuynh18: We used to have a file: `src/subscribers/request-subscriber.ts` that supposedly logged on every request update and creation. I didn't actually see it working even tho we did use it in the typeorm configs. We have an in-code log on every update so I added a log event for creation just in case we actually wanted those logs but they were broken before and no one noticed. I can remove this if its not neccessary

src/models/request.ts
stephhuynh18: I'll actually do this in a follow up PR

src/repositories/request-repository.ts
stephhuynh18: Yup will do and thanks for all the feedback. Its very helpful. I'll take a look at the PR and try to make it easier to review and remove refactors that are not necessary. I also have fresh eyes now which will help a lot in figuring out where I went too hard.

migrations/20220818164120_create_anchor_and_request.js
```diff
@@ -0,0 +1,102 @@
...
+
+const sortByStringKeys = (obj) =>
+  Object.fromEntries(Object.entries(obj).sort(([a], [b]) => a.localeCompare(b)))
+
+const checkObjEquality = (expected, received) =>
+  JSON.stringify(sortByStringKeys(expected)) === JSON.stringify(sortByStringKeys(received))
+
+const verifyTableStructure = async (knex, tableWithValidStructure, tableToCheck) => {
+  const [expectedStructure, existingStructure] = await Promise.all([
+    knex(tableWithValidStructure).columnInfo(),
```
willex, line 13: Instead of creating temp tables for comparison, you could just ship the column info dump as an object with the migration file.

migrations/20220818164120_create_anchor_and_request.js
willex: Yes, very similar in terms of implementation as well. Added the note about the temp table creation vs. shipping the table structure as an object.

**Minor nitpicks:**
- would be better to store with time zone so we enforce UTC
- instead of raw you can use the built in `now()` function

**Example:**
`table.dateTime('updated_at').notNullable().defaultTo(dbConnection.fn.now())`

results in:
`updated_at        | timestamp with time zone |           | not null | CURRENT_TIMESTAMP`

stephhuynh18: Yes i agree with these minor nitpicks. I did these weird things to keep as close to the current table structures as possible as to not interfere with anything. I ideally would like to use `timestamps` which includes a trigger. The interferences I can think of that may or may not appear is logging, metrics, related things. If these don't seem like a big deal i'll change things to timestamps and implement the shipping the table structure as an object

.env.prod.example
stephhuynh18: yes

stephhuynh18:
I believe these latest changes should be much easier to review. I moved back a bunch of things and made sure things stay in a similar order as the original. 

The biggest change will still be the `request-repository.ts`. The reason being is I took the opportunity to refactor the exceptionally complicated `findAndMarkReady` that i wrote before. I broke it down into steps that should (hopefully be much easier to understand). 

I also did some changes to the ceramic_integration test because it was failing 

I also am not going to do anything about the @willex 's nitpicks (sorry). This is because we have existing entries and changing the timezones is scary.

src/__tests__/test-utils.ts
```diff
@@ -139,29 +138,39 @@ export class MockEventProducerService implements EventProducerService {
   destroy(): void {}
 }
 
+export async function generateRequest(override: Partial<Request>) {
```
ukstv, line 141: Optional suggestion. Feel free to disregard. It might be better to create `IRequest` interface.

src/db-connection.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  return connection
+}
+
+/**
+ * USED FOR TESTING
+ * Clears all tables
+ * @param connection
+ */
+export async function clearTables(connection: Knex): Promise<void> {
```
ukstv, line 62: @stephhuynh18 Have you checked against Postgres? There are few slight surprising differences between SQLite and PG backends for Knex. One of them is `count` response, another is about table names, that I remember.

I am pretty sure you have, but it's worth stating that explicitly. I am just worried about some bug to appear out of the blue on Pg.

src/models/request.ts
```diff
@@ -43,3 +26,9 @@ export interface RequestUpdateFields {
 export const REQUEST_MESSAGES = {
   conflictResolutionRejection: 'Request has failed. Updated was rejected by conflict resolution.',
 }
+
+export class InvalidRequestStatusError extends Error {
```
ukstv, line 30: Feel free to disregard. It _might_ be a good idea to put the error to own file.

src/repositories/request-repository.ts
```diff
@@ -35,164 +31,215 @@ const countRetryMetrics = (requests: Request[], anchoringDeadline: Date): void =
...
+  return query
+}
+
+const findRequestsToAnchorForStreams = (
+  connection: Knex,
+  streamIds: string[],
+  now: Date
+): Knex.QueryBuilder =>
+  findRequestsToAnchor(connection, now)
+    .andWhere('streamId', 'in', streamIds)
```
ukstv, line 78: `whereIn`? Feel free to disregard though.

src/repositories/request-repository.ts
```diff
@@ -35,164 +31,215 @@ const countRetryMetrics = (requests: Request[], anchoringDeadline: Date): void =
...
-            type: 'request',
-            ...request,
-            ...fields,
-            createdAt: request.createdAt.getTime(),
-            updatedAt: new Date(Date.now()),
-          })
-        })
-        return result
+    const result = await connection(TABLE_NAME)
+      .update({ ...fields, updatedAt: updatedAt })
```
ukstv, line 143: This `...fields` thing could backfire IMO at some point by trying to update a non-existent property. Maybe pass the fields explicitly? I know it might be considered an unnecessary duplication, but I see the risk here.

Feel free to act on that as you wish though @stephhuynh18

.env.dev.example
```diff
@@ -13,12 +13,8 @@ ETH_RPC_HOST=http://localhost
...
-TYPEORM_CONNECTION = sqlite
-TYPEORM_HOST = localhost
-TYPEORM_USERNAME = root
-TYPEORM_PASSWORD = root
-TYPEORM_DATABASE = ./data/anchor.db.sq3
-TYPEORM_PORT = 5432
-TYPEORM_SYNCHRONIZE = true
-TYPEORM_LOGGING = true
-TYPEORM_ENTITIES = build/models/**/*.js
+DB_HOST = localhost
```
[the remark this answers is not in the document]
ukstv: Had a look at CircleCI workflow, I am just confused. We use `DATABASE_URL` on CircleCI but here we use full set of variables. Could you please tell why?

src/__tests__/test-utils.ts
ukstv: UPD: Maybe not, since Request is a pretty dumb class now.

src/db-connection.ts
ukstv: UPD: Please, disregard this.

src/repositories/request-repository.ts
```diff
@@ -35,164 +31,215 @@ const countRetryMetrics = (requests: Request[], anchoringDeadline: Date): void =
...
-          .from(Request, 'recent_request')
-          .orderBy('recent_request.updated_at', 'DESC')
-          .where('recent_request.updated_at >= :deadlineDate', {
-            deadlineDate: deadlineDate.toISOString(),
-          })
-          .getQuery()
-        return 'request.doc_id NOT IN ' + recentRequestsStreamIds
-      })
-      .getMany()
+    const recentlyUpdatedRequests = connection(TABLE_NAME)
```
stbrody, line 231: `requestsOnRecentlyUpdatedStreams`?
Something to make clear that we're not actually just skipping requests that were recently updated (which would already be handled by the `andWhere('updatedAt', '<', deadlineDate)`), but we also want to skip requests that haven't been updated recently directly, but where there are other requests that *were* updated recently on the same streamid

src/repositories/request-repository.ts
```diff
@@ -202,165 +249,88 @@ export class RequestRepository extends Repository<Request> {
...
-            createdAt: request['min_created_at'],
-          }))
-        )
+    const { connection = this.connection } = options
+    const now = new Date()
+    const anchoringDeadline = new Date(now.getTime() - this.config.maxAnchoringDelayMS)
+
+    return connection.transaction(
+      async (trx) => {
+        const streamsToAnchor = await findStreamsToAnchor(trx, maxStreamLimit, now)
```
stbrody, line 268: :heart_eyes: amazing cleanup

src/repositories/request-repository.ts
```diff
@@ -35,164 +31,215 @@ const countRetryMetrics = (requests: Request[], anchoringDeadline: Date): void =
   if (failed.length > 0) Metrics.count(METRIC_NAMES.RETRY_FAILED, failed.length)
 }
 
-@singleton()
-@EntityRepository(Request)
-export class RequestRepository extends Repository<Request> {
-  private readonly isolationLevel
+const findRequestsToAnchor = (connection: Knex, now: Date): Knex.QueryBuilder => {
```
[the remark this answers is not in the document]
stbrody: Looks like `findAndMarkReady` has a comment with a lot of what we want here. I'd take a lot of that text and move it here, then re-write the comment for findAndMarkReady to be more focused on what it means to select a group of streams and mark them READY - explaining how the interaction between the scheduler and the workers

src/repositories/request-repository.ts
```diff
@@ -202,165 +249,88 @@ export class RequestRepository extends Repository<Request> {
...
           return []
         }
 
         // Anchor if we have enough streams or the earliest stream request is expired
-        const earliestIsExpired = streamsToAnchor[0].createdAt < anchoringDeadline
-        if (streamsToAnchor.length >= minStreamLimit || earliestIsExpired) {
+        const enoughStreams = streamsToAnchor.length >= minStreamLimit
+        const earliestIsExpired = streamsToAnchor[0].minCreatedAt < anchoringDeadline
+
+        if (enoughStreams || earliestIsExpired) {
```
stbrody, line 280: Could we invert this condition and early return?  That would avoid an indentation layer and I think make this function easier to follow

src/repositories/request-repository.ts
```diff
@@ -35,164 +31,215 @@ const countRetryMetrics = (requests: Request[], anchoringDeadline: Date): void =
...
+    )
+    .orWhere((builder) =>
+      builder
+        .where({ status: RequestStatus.PROCESSING })
+        .andWhere('updatedAt', '<', processingDeadline)
+    )
+    .orWhere({ status: RequestStatus.PENDING })
+}
+
+const findStreamsToAnchor = (
```
stbrody, line 57: Should this just return an Array of StreamIDs?

src/repositories/request-repository.ts
```diff
@@ -202,165 +249,88 @@ export class RequestRepository extends Repository<Request> {
...
 
-          // using conditions in allRequestsToAnchorQuery, retrieves all requests associated with the retrieved streams
-          const requests = await allRequestsToAnchorQuery
-            .orderBy('request.createdAt', 'ASC')
-            .andWhere('request.streamId IN (:...streamIds)', { streamIds })
-            .getMany()
+          const requests = await findRequestsToAnchorForStreams(trx, streamIds, now)
 
-          const results = await this.updateRequests(
+          const updatedCount = await this.updateRequests(
```
stbrody, line 285: nitpick: consider a markRequestsReady helper that can handle the query plus checking if the updatedCount doesn't match the requests length

src/repositories/request-repository.ts
```diff
@@ -202,165 +249,88 @@ export class RequestRepository extends Repository<Request> {
...
+          `Not updating any requests to READY because there are not enough streams for a batch ${streamsToAnchor.length}/${minStreamLimit} and the earliest request is not expired (created at ${streamsToAnchor[0].minCreatedAt})`
         )
 
         return []
-      })
-      .catch(async (err) => {
-        if (err?.code === REPEATED_READ_SERIALIZATION_ERROR) {
-          Metrics.count(METRIC_NAMES.DB_SERIALIZATION_ERROR, 1)
-          await Utils.delay(100)
-          return this.findAndMarkReady(maxStreamLimit, minStreamLimit)
```
stbrody, line 306: what happened to this retry logic?  Is it not needed anymore?

stbrody (commented):
looking great.  Left some comments but they're all pretty small nitpicks around commenting and small code cleanups

src/repositories/request-repository.ts
ukstv: Hmm, that's true. Missed that type guard. It feels safer still to make the arguments explicit, but I am ok with leaving the code as it is.

.env.dev.example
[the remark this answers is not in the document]
stephhuynh18: In our tests we use "@databases/pg-test" which uses docker to launch a postgres db. This allows anyone to run all the tests locally without having to set up their own postgres. 

If DATABASE_URL is set, pg-test does nothing. So you can run tests locally using your own postgres instances by setting DATABASE_URL to point to it.

For use of pg-test with CircleCI, it was recommended via pg-test's docs to launch a postgres using CircleCI and then set DATABASE_URL so pg-test uses it.

>But, no, you could use single variable for connection string here, and pass it to the config file. That would be a breaking change, but it is possible still, and we are the only ones running CAS at the moment, so no one external is affected really IMO.

Yes that is true. I originally kept the connection as an object to minimize the typeorm --> knex changes BUT since all of our ENV vars are prefixed with TYPEORM_XXX we will have to make a bunch of changes anyway. I'll make this change.

src/repositories/request-repository.ts
stephhuynh18: good catch! i'll add that back in

.env.dev.example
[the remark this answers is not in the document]
ukstv: Oh, this is so very complicated. I am glad you own it.

migrations/20221004154924_create_anchor_and_request.js
```diff
@@ -0,0 +1,177 @@
...
+const createRequestTable = async (knex, tableName) => {
+  await knex.schema.createTable(tableName, (table) => {
+    table.uuid('id').notNullable().unique().primary().defaultTo(knex.raw('uuid_generate_v4()'))
+    table.integer('status').notNullable()
+    table.boolean('pinned').notNullable().defaultTo(false)
+    table.timestamp('created_at', { useTz: false }).notNullable().defaultTo(knex.raw('now()'))
+    table.timestamp('updated_at', { useTz: false }).notNullable().defaultTo(knex.raw('now()'))
+  })
+
+  await knex.schema
```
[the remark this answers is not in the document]
ukstv: That's a bummer.

stephhuynh18:
Gotta change env vars before this can be merged: https://github.com/3box/ceramic-infra/pull/638

src/repositories/request-repository.ts
```diff
@@ -263,21 +297,31 @@ export class RequestRepository {
     const now = new Date()
     const anchoringDeadline = new Date(now.getTime() - this.config.maxAnchoringDelayMS)
 
-    return connection.transaction(
-      async (trx) => {
-        const streamsToAnchor = await findStreamsToAnchor(trx, maxStreamLimit, now)
+    return connection
+      .transaction(
+        async (trx) => {
+          const streamsToAnchor = await findStreamsToAnchor(trx, maxStreamLimit, now)
```
stbrody, line 303: could all the logic from this line through line 325 be moved into `findStreamsToAnchor`?  Then findStreamsToAnchor could just return an array of StreamIDs, without needing to define the new `streamAndEarliestRequestDate` type, plus the logic in this function gets smaller and more straightforward

stbrody (approved):
lgtm with one optional suggested code cleanup

PR #759, opened 2022-09-20
fix: Use stream's state metadata to generate merkle tree metadata

stephhuynh18:
```
[2022-09-20T14:42:09.433Z] ERROR: 'Error when anchoring: Error: Merkle tree cannot be created: TypeError: candidate.metadata.controllers is not iterable' Will look into fixing
```

No tests because that is a larger effort (adding to the ceramic_integration repo and setting up all the model stuff)

### DIFF
--- a/src/merkle/merkle-objects.ts
+++ b/src/merkle/merkle-objects.ts
@@ -157,7 +157,7 @@ export class Candidate implements CIDHolder {
       this._alreadyAnchored = true
     } else {
       this._cid = stream.tip
-      this._metadata = stream.metadata as StreamMetadata
+      this._metadata = stream.state.metadata as StreamMetadata
     }
 
     // Check the log of the Stream that was loaded from Ceramic to see which of the pending requests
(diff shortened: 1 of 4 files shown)


src/merkle/merkle-objects.ts
```diff
@@ -157,7 +157,7 @@ export class Candidate implements CIDHolder {
       this._alreadyAnchored = true
     } else {
       this._cid = stream.tip
-      this._metadata = stream.metadata as StreamMetadata
+      this._metadata = stream.state.metadata as StreamMetadata
```
stbrody, line 160:
```suggestion
      this._metadata = stream.state.next?.metadata ? stream.state.next.metadata : stream.state.metadata as StreamMetadata
```

stbrody: sigh.  Just checking the state.metadata would work for Models and MIDs, but for TileDocument we need to also check for state.next.metadata since TileDocument does actually allow changing controllers.

stbrody (approved):
lgtm mod making sure to handle state.next for TileDocument streams

PR #760, opened 2022-09-21
fix: use min stream count when marking requests as ready using the scheduler

stephhuynh18:
Originally we wanted the min stream count = max stream count aka only anchor when we have a 1024 streams or streams are expiring.

This means that we will have to set MIN_STREAM_COUNT to something in dev and prod otherwise it will default to 1 (default.json)


stbrody (approved):
lgtm

PR #761, opened 2022-09-23
feat: add ready retry interval to config

v-stickykeys:
# Ready Retry Interval

## Description

Allows us to configure the interval for retrying requests marked as ready. "Retrying" means we emit an anchor event that these requests should be processed by. (Note: This is for a case in which a request marked as ready has not gotten processed yet, which can happen for example if the currently running workers are at capacity. So the next worker that starts will know to pick up the retried request if it can.)

## How Has This Been Tested?

TODO

## Notes

Related to an infra change https://github.com/3box/ceramic-infra/pull/634


README.md
```diff
@@ -270,6 +270,12 @@ In order to run tests with coverage run:
 
 ```npm run coverage```
 
+## Contributing
+
+### Config
+
+Values in the config files get their types auto-generated by [node-config-ts](https://github.com/tusharmath/node-config-ts) when running `npm run postinstall`.
+
 ## Maintainers
```
smrz2001, line 279: Can we finally change this lol? It bugs me everytime I come across it. I'd say @stbrody and @stephhuynh18 should probably be here instead.

stephhuynh18 (approved):
LGTM but i have no idea why the code coverage is failing

README.md
v-stickykeys: haha yeah I will finalize a pr with changes for this

v-stickykeys: see https://github.com/ceramicnetwork/ceramic-anchor-service/pull/711

PR #774, opened 2022-10-07
chore(api): print full stack trace for api catch-all exceptions

smrz2001:


stephhuynh18 (approved):

src/controllers/request-controller.ts
```diff
@@ -113,7 +113,7 @@ export class RequestController {
       }
     } catch (err) {
       const errmsg = `Creating request with streamId ${req.body.streamId} and commit CID ${req.body.cid} failed: ${err.message}`
-      logger.err(errmsg)
+      logger.err(err)
```
stbrody, line 116: this change loses information (the streamid and commit CID that originated the error).
I put up https://github.com/ceramicnetwork/ceramic-anchor-service/pull/776 to log both the original message and also the stack the trace

smrz2001: Ah good, point. Ty 🙏🏼

PR #784, opened 2022-10-12
add performance traces and additional cas metrics

gvelez17:
# performance traces & other metrics - #PLAT-844, #PLAT-1024

## Description

We need to identify performance bottlenecks, tracing is a standard way to do this

Also we want metrics on pin success & failure

## How Has This Been Tested?

- [x] added unit test 
- [x] tested with exporter configured on dev (deployed branch) to ensure expected metrics show up

## Definition of Done

- [x] The changes have been communicated to interested parties

## References:


src/__tests__/service-metrics.test.ts
```diff
@@ -4,6 +4,12 @@ describe('simple test of metrics', () => {
   beforeAll(async () => {
     ServiceMetrics.start()
   })
+  test('trace span', async() => {
```
ukstv, line 7: What does the test check? o_O If it is an intentional noop to check availability of methods, could you mind please add a comment explaining that? I am afraid some next person could remove this test.

src/service-metrics.ts
```diff
@@ -3,55 +3,137 @@
...
+    this.tracer = trace.getTracer(caller)
+
+    // accept a logger from the caller
+    this.logger = logger || new _NullLogger()
   }
 
   // could have subclasses or specific functions with set params, but we want to
   // easily and quickly change what is recorded, there are no code dependencies on it
 
+  start_span(name: string, params?: any) {
```
ukstv, line 120: Any reason for not using camelCase?

ukstv: Also, please, add return types.

src/service-metrics.ts
```diff
@@ -3,55 +3,137 @@
...
+
+enum LOG_LEVEL {
+    debug,
+    info,
+    warn,
+    imp,
+    error
+}
+
+class _NullLogger {
```
ukstv, line 29: Why leading underscores here?

src/settings.ts
```diff
@@ -2,12 +2,21 @@ export enum METRIC_NAMES {
   ANCHOR_SUCCESS = 'anchor_success',
   ANCHOR_TIMEOUT = 'anchor_timeout',
```
ukstv, line 3: It seems, the file is only metrics related. Maybe reflect that in name, rather than use generic `settings`?

ukstv (changes requested):

src/service-metrics.ts
gvelez17: this is an internal method only - i can remove

src/settings.ts
ukstv: IDK, I see no other settings for now, so your proposal for renaming makes all sense to me. IMO, it makes sense to use dash instead of underscore also, to be uniform with other places in code.

src/service-metrics.ts
gvelez17: changed, and added!

src/__tests__/service-metrics.test.ts
gvelez17: yeah, the metrics methods are all no-op unless an exporter is actually present.  added comment

src/__tests__/service-metrics.test.ts
```diff
@@ -4,6 +4,12 @@ describe('simple test of metrics', () => {
...
+    const span = ServiceMetrics.start_span("doing it")
```
ukstv:
```suggestion
    // We only check here if we *can call* the methods
    const span = ServiceMetrics.start_span("doing it")
```

Maybe something like that to make the intent explicit?

gvelez17: sure, done.

ukstv (approved):
Yay! All right, I can live with `_NullLogger`, I think.

PR #788, opened 2022-10-21
ci: rename test workflow

v-stickykeys:
Just rename the test workflow so that it is more obvious what is running in the github ui


smrz2001 (approved):
🚀

PR #794, opened 2022-10-28
fix: serializable isolation lvl and already anchored check

stephhuynh18:
Use serializable isolation level to prevent concurrent updates from making unpredictable updates

Check if the newest request was already anchored, if so mark all those requests as anchored


v-stickykeys (approved):
🛳

PR #795, opened 2022-10-28
fix: serializable isolation lvl and already anchored check

stephhuynh18:
Use serializable isolation level to prevent concurrent updates from making unpredictable updates

Check if the newest request was already anchored, if so mark all those requests as anchored

### DIFF
--- a/src/services/__tests__/anchor-service.test.ts
+++ b/src/services/__tests__/anchor-service.test.ts
@@ -22,7 +22,7 @@ import type { Knex } from 'knex'
 import { CID } from 'multiformats/cid'
 import { Candidate } from '../../merkle/merkle-objects.js'
 import { Anchor } from '../../models/anchor.js'
-import { AnchorStatus, toCID } from '@ceramicnetwork/common'
+import { AnchorStatus, LogEntry, CommitType, toCID } from '@ceramicnetwork/common'
 import cloneDeep from 'lodash.clonedeep'
 import { Utils } from '../../utils.js'
 import { PubsubMessage } from '@ceramicnetwork/core'
@@ -69,14 +69,19 @@ async function anchorCandidates(
   await anchorService._persistAnchorResult(anchors, candidates)
   return anchors
 }
-
-function createStream(id: StreamID, log: CID[], anchorStatus: AnchorStatus = AnchorStatus.PENDING) {
+function createStream(
+  id: StreamID,
+  log: CID[] | LogEntry[],
+  anchorStatus: AnchorStatus = AnchorStatus.PENDING
+) {
   return {
     id,
     metadata: { controllers: ['this is totally a did'] },
     state: {
-      log: log.map((cid) => {
-        return { cid }
+      log: log.map((logEntry) => {
+        const cid = CID.asCID(logEntry)
+        if (cid) return { cid }
+        return logEntry
       }),
       anchorStatus,
       metadata: { controllers: ['this is totally a did'] },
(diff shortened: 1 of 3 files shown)


src/services/__tests__/anchor-service.test.ts
```diff
@@ -715,6 +720,70 @@ describe('anchor service', () => {
       const updatedRequest = await requestRepository.findByCid(toCID(request.cid))
       expect(updatedRequest.status).toEqual(RequestStatus.COMPLETED)
     })
+
+    test('Request succeeds without anchor for already CIDs with next CIDs that have been anchored', async () => {
```
stbrody, line 724: grammar of this test description seems off, I don't really understand what it's trying to say?

stbrody (approved):
do we know how we could get into this situation in the first place?  Why wouldn't the batch that anchored the later request have already marked the earlier requests as completed?

src/services/__tests__/anchor-service.test.ts
stephhuynh18: i'll change this another PR because integration tests take too long. Fuzzy brain did this word dump haha. What i really mean to say is: 

Request succeeds without anchor if subsequent CIDs are already anchored

stephhuynh18:
> do we know how we could get into this situation in the first place? Why wouldn't the batch that anchored the later request have already marked the earlier requests as completed?

Yes this is exactly what i thought too. 
RequestA for CID1 for StreamX has already been anchored but its status does not reflect this and is PENDING
StreamX has a new tip (CID2)
When RequestA is picked up in the next batch. StreamX is now an anchor candidate. StreamX's anchor state is PENDING because of it's new tip CID2. So an anchor is created for StreamX. Unfortunately the anchor commit created is associated with RequestA causing a duplicate key error, even though the anchor commit is valid.

stbrody:
> RequestA for CID1 for StreamX has already been anchored but its status does not reflect this and is PENDING

So, this could happen given the problem we've been having, where we've been having errors marking streams as anchored causing that whole step to fail leaving some streams in PENDING that were actually anchored.  But this is the reason why we think this is happening in the first place!  How did the *first* stream get into this state?

stephhuynh18:
> > RequestA for CID1 for StreamX has already been anchored but its status does not reflect this and is PENDING
> 
> So, this could happen given the problem we've been having, where we've been having errors marking streams as anchored causing that whole step to fail leaving some streams in PENDING that were actually anchored. But this is the reason why we think this is happening in the first place! How did the _first_ stream get into this state?


YES exactly. So in the chat i actually have two pictures of logs which show the same request being updated to COMPLETED by an anchor worker at 12:15:45 and then updated to READY at 23:15:51 by the scheduler service. How did this happen when the isolation levels were set to `repeatable read`....

stephhuynh18:
> > RequestA for CID1 for StreamX has already been anchored but its status does not reflect this and is PENDING
> 
> So, this could happen given the problem we've been having, where we've been having errors marking streams as anchored causing that whole step to fail leaving some streams in PENDING that were actually anchored. But this is the reason why we think this is happening in the first place! How did the _first_ stream get into this state?

Misread a bit but what i said above echoes your last question. How did this _first_ stream get into this state? (how did these logs come to be...)

For the whole PENDING bit. Anchors that have unexpected errors are reverted back to PENDING. That is correct. BUT their anchor commits are not persisted in the DB (unless something happened between persisting the anchor commits and exiting which is highly unlikely but i guess not unheard of. I do see the anchor completed successfully tho)

PR #801, opened 2022-10-31
fix: use db transaction when updating expiring ready requests

stephhuynh18:
When updating expiring ready requests we were not using a db transaction. 

1a. Scheduler service gathers expired ready requests which include requestA
1b. requestA is at the SAME time being updated to PENDING and added to a batch
2. all requests gathered in step 1a are updated to READY via the scheduler service
requestA should have status PENDING but because of step 2 it is READY

3a. The scheduler service gathers expired ready requests and thinks requestA is READY and expired
3b. The batch completes and requestA is at the SAME time being updated to COMPLETED
4. All requests gathered in step 3a are updated to READY via the scheduler service
requestA should have status COMPLETED but because of step 4 it is READY

Not ideal at all so I put steps 1a + 2 into a db transaction to prevent the updates using old row data

### DIFF
--- a/src/repositories/request-repository.ts
+++ b/src/repositories/request-repository.ts
@@ -387,8 +387,8 @@ export class RequestRepository {
   }
 
   /**
-   * Finds and updates all READY requests that have not been moved to PROCESSING in a sufficient amount of time
-   * Updates them again to indicate that they are being retried
+   * Finds and updates all READY requests that are expired (have not been moved to PROCESSING in a sufficient amount of time)
+   * Updating them indicates that they are being retried
    * @param options
    * @returns A promise for the number of expired ready requests updated
    */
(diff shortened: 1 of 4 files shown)


src/repositories/request-repository.ts
```diff
@@ -385,4 +385,53 @@ export class RequestRepository {
...
+          )
+
+          return updatedCount
+        },
+        {
+          isolationLevel: 'serializable',
+        }
+      )
+      .catch(async (err) => {
+        if (err?.code === REPEATED_READ_SERIALIZATION_ERROR) {
```
stbrody, line 428: is this still the code we should be checking for when using `serializable` as the isolation level?

stbrody (approved):

src/repositories/request-repository.ts
stephhuynh18: Good call out!

Yes as referenced here: https://www.postgresql.org/docs/15/mvcc-serialization-failure-handling.html

PR #803, opened 2022-11-01
feat: Add indexes

stephhuynh18:
We do not have any indexes in our database. Hopefully this improves performance

### DIFF
--- a/migrations/20221031220419_alter_request_add_indexes.js
+++ b/migrations/20221031220419_alter_request_add_indexes.js
@@ -1,4 +1,5 @@
 const REQUEST_TABLE_NAME = 'request'
+const ANCHOR_TABLE_NAME = 'anchor'
 
 /**
  * @param { import("knex").Knex } knex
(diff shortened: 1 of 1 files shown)


migrations/20221031220419_alter_request_add_indexes.js
```diff
@@ -0,0 +1,31 @@
+const REQUEST_TABLE_NAME = 'request'
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex) {
```
stbrody, line 7: what about indexes for the anchor table?  We query anchors by requestId so should probably have an index on requestId.  Also should probably have one on anchor cid while we're at it

migrations/20221031220419_alter_request_add_indexes.js
```diff
@@ -0,0 +1,31 @@
...
+  await knex.schema.alterTable(REQUEST_TABLE_NAME, (table) => {
+    table.index(['stream_id'])
+    table.index(['status'])
```
stbrody, line 10: We shouldn't need this since we have the multicolumn indexes that start with status below

migrations/20221031220419_alter_request_add_indexes.js
stephhuynh18: we don't have any queries using anchor cid, but i'll add one for requestId

stephhuynh18: also we do not need one for requestId because this column is unique. Postgres auto makes unique indexes for primary keys and unique columns. 

The note in their docs: There's no need to manually create indexes on unique columns; doing so would just duplicate the automatically-created index (https://www.postgresql.org/docs/current/indexes-unique.html)

smrz2001 (approved):
🚀

PR #805, opened 2022-11-03
chore: use repeatable read instead of serializable in db transactions

stephhuynh18:
This was not the actual cause of our race condition issues


stbrody:
> This was not the actual cause of our race condition issues

What was?

stbrody (approved):

stephhuynh18:
> > This was not the actual cause of our race condition issues
> 
> What was?

this PR https://github.com/ceramicnetwork/ceramic-anchor-service/pull/801

PR #806, opened 2022-11-03
feat: add pendingRequestCount to anchor summary

stbrody:


gvelez17 (approved):
great, this will be very helpful - I assume we should only call this periodically ?

src/repositories/request-repository.ts
```diff
@@ -386,6 +386,18 @@ export class RequestRepository {
     return query
   }
 
+  /**
+   * Returns the number of pending anchor requests that remain in the database.
+   * @returns The number of requests in the database in status PENDING
+   */
+  public async countPendingRequests(): Promise<number> {
```
gvelez17, line 393: should we put a note that this is a db call and should not be made too frequently?  Or we assume people know that

stbrody: This whole RequestRepository class exists solely to contain the interactions with the request table in the database, so I think it's sufficiently implied

stbrody:
> great, this will be very helpful - I assume we should only call this periodically ?

This will get called once at the end of every anchor batch

stephhuynh18 (approved):

PR #810, opened 2022-11-08
feat: upsert anchor commits

stephhuynh18:
Prevents the duplicate key error but we need to set an alarm for the reanchored metric


src/repositories/anchor-repository.ts
```diff
@@ -21,7 +24,19 @@ export class AnchorRepository {
    */
   public async createAnchors(anchors: Array<Anchor>, options: Options = {}): Promise<void> {
     const { connection = this.connection } = options
-    await connection(TABLE_NAME).insert(anchors)
+
+    const result = (await connection
+      .table(TABLE_NAME)
+      .insert(anchors)
+      .onConflict('requestId')
```
stbrody, line 31: I think we need to make sure that more of the anchor is the same before we ignore conflicts.  They should probably have at least the same `requestId`, `cid`, and `proofCid` I think

stbrody: Can you remind me what the sequence of events is that can lead to conflicts here that we're trying to handle?  Are these duplicate key errors coming from the same commit getting anchored into two different anchor batches, with different merkle trees? If so then I guess the proofCid would be different.  Seems like that shouldn't happen though?

stbrody (changes requested):

src/repositories/anchor-repository.ts
stephhuynh18: the `cid` and `proofCid` will not be the same as the stream was anchored again. In this case we are preferring the already existing anchor as conflict resolution will prefer it the anchor commit that references it.

stephhuynh18: We have BatchA that contains RequestX that is taking a >3 hours to process

Scheduler service sees RequestX has been `processing` for >3 hours. It marks it as READY. 

BatchB starts with RequestX. 

BatchA finishes first. It creates a row in the anchor table for RequestX
BatchB finishes and tries to create a row in the anchor table for RequestX. It fails because of a duplicate key error. 

RequestX has been anchored successfully twice (in two different eth txs) with 2 anchor commits created and published. The earliest anchor commit is preferred so we `ignore` conflicts when inserting new anchors in the anchor table

stephhuynh18: I would say this is the workaround, whereas the real problem is the scheduler service not knowing that RequestX is being ACTIVELY worked on. The scheduler service should not add RequestX to a batch if so.  I am not sure how the scheduler service can know that.

stbrody: ah yeah, we shouldn't move a request out of PROCESSING back into READY unless we're confident that the worker that had been handling the processing request is *actually* dead and definitely won't be coming back.  @smrz2001 any thoughts on how we can synchronize the timeout for requests that have been in PROCESSING for too long with knowledge of when the job that is supposed to be processing them has been killed?

stbrody:
Seems like for this PR, maybe we don't actually want to upsert the anchors as a conflict here indicates a real issue, but maybe we should just log and swallow the error rather than failing the whole batch?  We should also raise a metric so we can alert on it.

stephhuynh18:
> Seems like for this PR, maybe we don't actually want to upsert the anchors as a conflict here indicates a real issue, but maybe we should just log and swallow the error rather than failing the whole batch? We should also raise a metric so we can alert on it.

yup the title of the PR doesn't actually reflect what is being done which is creating the anchors but ignoring the duplicates. It is not upserting. Since we are inserting all the anchors in one query I can't let it error but I do log the number of anchors that were duplicates. I will change the title of the PR.

src/services/anchor-service.ts
```diff
@@ -225,17 +227,29 @@ export class AnchorService {
 
     // Update the database to record the successful anchors
     logger.debug('Persisting results to local database')
-    const numAnchoredRequests = await this._persistAnchorResult(anchors, candidates)
+    const persistedAnchorsCount = await this._persistAnchorResult(anchors, candidates)
```
stbrody, line 230: its getting a little hard to keep track of the difference between `anchorCount`, `anchoredRequestCount`, and `reAnchoredCount` (and how they correspond to `anchors` and `anchoredRequests`).  Could we add some comments to make that clearer?

stbrody (approved):
lgtm mod adding some comments

PR #822, opened 2022-11-21
chore: Version bumps

ukstv:
- Ordinary version bumps
- Conformance with lru-cache types
- Do not depend on @ipld/dag-cbor package. We use only single codec identifier from there, and just for test purposes.


ukstv:
Please, disregard image verification failure. It is fixed in https://github.com/ceramicnetwork/ceramic-anchor-service/pull/828 which eventually should land here.

pawartur (approved):

PR #824, opened 2022-11-22
Pin merkle metadata object

PaulLeCam:
## How Has This Been Tested?

- [x] Added unit tests for `storeRecord()` method
- [x] Added check for internal API calls

## Definition of Done


PaulLeCam:
@oed @stbrody looks like the Ceramic integration tests are failing due to the timeout when attempting to store records in IPFS.
Should I increase the timeout or revert to only pinning the metadata please?

stbrody:
Pretty surprising it takes over 30 seconds just to store and pin a record.  I guess try bumping the timeout to one minute and see if that helps?

src/services/ipfs-service.ts
```diff
@@ -116,11 +117,16 @@ export class IpfsServiceImpl implements IpfsService {
   /**
    * Sets the record and returns its CID
    * @param record - Record value
+   * @param pin - Add to the pin store
    */
-  public async storeRecord(record: Record<string, unknown>): Promise<CID> {
+  public async storeRecord(record: Record<string, unknown>, pin = true): Promise<CID> {
```
stbrody, line 122: why even have this argument if we never set it to false?

PaulLeCam: I made it opt-in to store the metadata only, then switched this argument to pin by default.
I changed the logic to always pin, without argument.

src/services/ipfs-service.ts
```diff
@@ -16,7 +16,7 @@ const { serialize, MsgType } = PubsubMessage
 
 const DEFAULT_GET_TIMEOUT = 30000 // 30 seconds
 const MAX_CACHE_ENTRIES = 100
-const IPFS_PUT_TIMEOUT = 60 * 1000 // 60 seconds
+const IPFS_PUT_TIMEOUT = 120 * 1000 // 2 minutes
```
stbrody, line 19: do you have any idea *why* it's taking 2 full minutes just to put a record into IPFS?

I'm actually kind of confused on what the expected latency is from ipfs puts.  I guess I would expect it to be comparable to reads, since you need to walk the DHT to find the peers with the closest XOR distance to the CID to register with them as providers.  But empirically, I've generally seen IPFS puts be extremely fast, far faster than gets.  Any idea what the expected behavior here is?  @oed @smrz2001 @zachferland?

oed: This is very strange yes. the "dag put" api should not wait for the DHT providing to finish afaik.

oed: If you try doing `echo '{ "test": 123 }' | ipfs dag put` on a local ipfs node it is instant!

stbrody: I guess the core of this change is adding a call to `pin` so I guess the real extra time is coming from the pin not the put, but that makes even less sense to take a long time to me.  I'd think pinning a CID is just making a small update to the local node's data files to mark that it shouldn't be garbage collected.  Should be super fast

smrz2001: Is this much time consistently needed locally and in tests as well, or was this updated because of something we observed on our infra?

On infra, it could be a variety of reasons, so I wouldn't update the default to deal with that.

oed: Doing `echo '{ "test": 123 }' | ipfs dag put --pin` is also instant on my machine.

oed: oh wait, recursive pins?
Not sure if that's enabled by default.

PaulLeCam: I haven't had timeout issues locally, only in CI, but now I have another issue preventing from running tests locally:
```
 src/app.ts:24:25 - error TS2307: Cannot find module './version.js' or its corresponding type declarations.

 24 import { version } from './version.js'
```

PaulLeCam: Other test suites run but I get `TypeError: fetch failed` with the `ceramic_integration` test locally so I can't check if the timeouts happen locally as well.

nathanielc (approved):

stbrody (approved):
LGTM

PR #827, opened 2022-11-25
Su/version bumps jest

ukstv:

## Description

## How Has This Been Tested?

- [ ] Test B

## Definition of Done

## References:


src/app.ts
```diff
@@ -165,8 +160,8 @@ export class CeramicAnchorApp {
    * @private
    */
   private async _startScheduler(): Promise<void> {
```
stbrody, line 162: Feels like we might want at least the healthcheck controller for the scheduler service?  At least let's make sure to confirm with platform that we aren't already depending on it.

stbrody (commented):
this is great, thanks for taking care of this Sergey!

> Found and fixed a bug with RequestRepository do not really merging requests with the same cid (see [this line](https://github.com/ceramicnetwork/ceramic-anchor-service/pull/827/files#diff-a46b9c12f9f4fa321c38a22c9b78928be28abe531cef9c6365186adb54d9fde0R170))

My one request is since this PR is so big, can we pull out the functional change from the refactoring?  Can we have a separate small PR just for the bug fix so we can keep this PR focused?  Also, if for some reason this switch from tsyringe to typed-inject caused something to break after merge, it would be good to be able to revert that one commit without also reverting the bug fix.

ukstv:
> My one request is since this PR is so big, can we pull out the functional change from the refactoring? Can we have a separate small PR just for the bug fix so we can keep this PR focused? Also, if for some reason this switch from tsyringe to typed-inject caused something to break after merge, it would be good to be able to revert that one commit without also reverting the bug fix.

Makes sense, there is one PR for the bug fix: https://github.com/ceramicnetwork/ceramic-anchor-service/pull/843

src/app.ts
ukstv: Do not quite understand this comment. The PR does not change the relationship between the services (apart removing some clearly unused dependency in some class).

src/app.ts
```diff
@@ -23,61 +24,56 @@ import { EthereumBlockchainService } from './services/blockchain/ethereum/ethere
...
-      ),
-    })
-    container.registerSingleton('eventProducerService', HTTPEventProducerService)
-    container.registerSingleton('anchorService', AnchorService)
-    container.registerSingleton('ceramicService', CeramicServiceImpl)
-    container.registerSingleton('ipfsService', IpfsServiceImpl)
-    container.registerSingleton('schedulerService', SchedulerService)
-
-    // register controllers
-    container.registerSingleton('healthcheckController', HealthcheckController)
```
stbrody, line 71: my comment about the healthcheck controller is due to this change.  We used to always register this controller, now we only register it if we call `_startServer`

src/app.ts
stbrody: We used to always register the healthcheck controller, including for the scheduler, but now we don't. Unless I'm missing where the HealthcheckController gets registered when running the scheduler?

ukstv: Okay, now I understand. I think, that nothing has changed with regards to how controllers start. Registration in a DI container is no more than a promise of future initialization. For greater confidence, going to create a separate branch to see if the controller in question is ever started in the scheduler mode.

stbrody: okay sounds good, I also saw your discord message confirming that platform isn't using it currently, so seems like it should be fine either way

stbrody:
@ukstv note that image verification failed in CI

stbrody (approved):
lgtm so long as there's no major changes needed to make CI pass

PR #842, opened 2022-11-28
feat: Let load stream timeout be configurable

stbrody:
The goal is that we can set this down on clay (probably to 1 second) so that Clay batches won't take so long to complete (given that we get a lot of stream load timeouts on clay, where it is more expected)


stbrody:
FYI @smrz2001

smrz2001 (approved):
🥳

stephhuynh18 (approved):

PR #844, opened 2022-11-29
feat(ci): use multi-stage docker build

smrz2001:
# Use multi-stage Docker build

## Description

Currently, we have two separate Docker files because of which I had been unable to build/push the images through Dagger (I couldn't figure out how to use a derived Dockerfile).

By implementing @3benbox's suggestion to use multi-stage builds, I was able to run the image builds as part of the Dagger image verification process.

This PR must be merged _before_ the `pipeline-tools` [PR](https://github.com/3box/pipeline-tools/pull/38) to use a multi-stage build for CAS, which also updates the CAS tests to use a Postgres container directly instead of using Docker Compose. The latter was causing image verification to fail intermittently due to container creation timing issues.

## How Has This Been Tested?

- [x] Ran the Dagger CI workflow with successful image build and verification 

## Definition of Done

- [x] I have updated the READMEs of affected packages
- [x] I have made corresponding changes to the documentation
- [x] The changes have been communicated to interested parties


.github/workflows/build_push_publish.yml
```diff
@@ -1,11 +1,12 @@
...
-  push:
-    branches:
-      - main
-      - "release-candidate"
-      - develop
+#  push:
+#    branches:
+#      - main
+#      - "release-candidate"
+#      - develop
```
3benbox, line 9: We're commenting these out because we don't want this new workflow to run yet?

3benbox (approved):
LGTM!

.github/workflows/build_push_publish.yml
smrz2001: This is actually the existing workflow that my changes will now make redundant. Didn't want to delete it yet, so I just disabled the automatic execution.

PR #853, opened 2022-12-06
feat: Add timestamp from input

ukstv:
Apparently, when PG driver in JS land gets timestamp from PG as text, it uses `new Date`. That uses _local_ timezone instead of UTC.

I wanted to test if passed timestamp is the same as in the database. Due to this local time zone usage, the returned value was different. Here I had to modify settings of PG driver so that "timestamp without timezone" is treated as UTC no matter what. Also, every timestamp passed in a database query has to be explicitly turned into UTC: `date.toISOString()` returns ISO8601 timestamp in UTC.

Also:
- Added `.editorconfig`. Minor stuff.
- Turned RequestPresentation into DI-managed RequestPresentationService. It really is a service. 
- Only accept `request.streamId`. Drop support of `request.docId`.


src/normalize-config.util.ts
```diff
@@ -4,7 +4,7 @@ import cloneDeep from 'lodash.clonedeep'
 function replaceBools(o: object): void {
   for (const prop of Object.keys(o)) {
     if (o[prop] === 'true' || o[prop] === 'false') {
-      o[prop] = (o[prop] === 'true')
+      o[prop] = o[prop] === 'true'
```
ukstv, line 7: Prettier.

pawartur (approved):

pawartur (changes requested):
Not really requesting changes, but I think it would be great, if we had a test that shows that without the custom parser, dates passed as strings to the db connection are stored with the local timezone.

src/repositories/request-repository.ts
```diff
@@ -170,7 +180,7 @@ export class RequestRepository {
     const keys = Object.keys(request).filter((key) => key !== 'id') // all keys except ID
     const [{ id }] = await connection
       .table(TABLE_NAME)
-      .insert(request, ['id'])
+      .insert(request.toDB(), ['id'])
```
stbrody, line 183: hmm... this scares me a little.  What happens if a developer fails to call `.toDB()` - will something clearly error?  Or will it silently insert the wrong data?

stbrody (approved):
lgtm mod one question about the new `toDB()` call on Requests.  Just want to confirm that there are some guardrails to make sure devs use it correctly.

src/repositories/request-repository.ts
ukstv: It will silently insert the wrong data, and it's okay. Someone has to explicitly convert runtime format to a database one. I am open to suggestions though.

I am generally not happy about runtime <-> db conversion. See no good solution for now, but it does not mean the search is over.

stbrody: ooph yeah that's a bummer.  Is there maybe a way we could configure the database to yell at us if the wrong type is inserted?  Like, if the date fields are inserted as Date objects, what do they show up as in the db?  Are they coerced to strings somehow or do they go in as numbers?  Can we make it so the db transaction would fail in that case?

It's weird that this seems to maybe be one place where TypeOrm was better than knex - TypeOrm has a way to specify custom serializer/deserializers for types, but my googling isn't finding anything equivalent for knex 😭

src/models/request.ts
```diff
@@ -15,6 +28,37 @@ export class Request {
...
+    this.streamId = params.streamId
+    this.message = params.message
+    this.pinned = params.pinned
+    this.createdAt = params.createdAt
+    this.updatedAt = params.updatedAt
+    this.timestamp = params.timestamp
+    this.origin = params.origin
+  }
+
+  toDB(): IDBRequest {
```
stbrody, line 48: Can we add a comment what this is doing and why it's necessary, with a big disclaimer making clear that you *have* to use this before inserting a request into the database or else your data will be silently corrupted?

stbrody (changes requested):
given the lack of any either compile-time or run-time checks that Request serialization is happening properly, I agree with Artur that we really should at least have a test that demonstrates this behavior and would catch a regression in the future

ukstv:
@stbrody @pawartur Here we had two contentious issues:
1. Some test code would be good to demonstrate how poorly Knex+PG+JS together handle timezones.
  I believe the issue is addressed here as part of the PR.
3. `toDB` of `Request` class: make a warning comment or do something better.
  I think, I have found a good enough way to deal with the serde. See [src/models/metadata.ts](https://github.com/ceramicnetwork/ceramic-anchor-service/pull/872/files#diff-59b8752a1d1722e73381bb3a414c384fe90d7749a98e681cae87ad1653432a88) in PR #872 and tell me if you are both okay with the approach. Then I would rather change `Request#toDB` implementation to make it similar to metadata serde.

stbrody:
> See [src/models/metadata.ts](https://github.com/ceramicnetwork/ceramic-anchor-service/pull/872/files#diff-59b8752a1d1722e73381bb3a414c384fe90d7749a98e681cae87ad1653432a88) in PR https://github.com/ceramicnetwork/ceramic-anchor-service/pull/872 and tell me if you are both okay with the approach. Then I would rather change Request#toDB implementation to make it similar to metadata serde.

Would this solve the issue of devs needing to remember to call toDb() though?

ukstv:
> > See [src/models/metadata.ts](https://github.com/ceramicnetwork/ceramic-anchor-service/pull/872/files#diff-59b8752a1d1722e73381bb3a414c384fe90d7749a98e681cae87ad1653432a88) in PR #872 and tell me if you are both okay with the approach. Then I would rather change Request#toDB implementation to make it similar to metadata serde.
> 
> Would this solve the issue of devs needing to remember to call toDb() though?

It will turn into more generic "use codec _every time_ when passing stuff to a database".

stbrody:
> It will turn into more generic "use codec _every time_ when passing stuff to a database".

But still requires devs to remember to do something special or else the data will be silently corrupted right?

Between the two then I don't feel strongly, happy to go with whichever you think is better.

ukstv:
> But still requires devs to remember to do something special or else the data will be silently corrupted right?

Unfortunately, one can not avoid or automate that.

PR #854, opened 2022-12-06
fix: get around postgres parameterized query limit

stephhuynh18:
Debated between two approaches: 
1. Limit amount of requests per stream so that the total amount of requests never exceeds some number. This prevents us from having to retrieve 65654 commits as well. The problem then shifts to: if we anchor 10 of the oldest requests for streamA, the next batch will pick up 10 more requests for streamA but streamA will be already anchored, so streamA will not be included in the batch and this will result in a possible emptier batch

2. I can circumvent this limit by using unnest as mohsin suggested in the comments of the discord message. This definitely works. But do we want to create a batch with >65654 requests (but 1024 streams)

I went with options 2. It is unlikely we will reach 65654 requests again, AND with the CAS w/o a ceramic node project, the loading problem does not exist.

### DIFF
--- a/src/repositories/request-repository.ts
+++ b/src/repositories/request-repository.ts
@@ -6,7 +6,11 @@ import { logEvent } from '../logger/index.js'
 import { Config } from 'node-config-ts'
 import { logger } from '../logger/index.js'
 import { Utils } from '../utils.js'
-import { ServiceMetrics as Metrics, TimeableMetric, SinceField } from '@ceramicnetwork/observability'
+import {
+  ServiceMetrics as Metrics,
+  TimeableMetric,
+  SinceField,
+} from '@ceramicnetwork/observability'
 import { METRIC_NAMES } from '../settings.js'
 
 // How long we should keep recently anchored streams pinned on our local Ceramic node, to keep the
@@ -221,7 +225,7 @@ export class RequestRepository {
         pinned: fields.pinned,
         updatedAt: updatedAt,
       })
-      .whereIn('id', ids)
+      .where('id', 'in', connection.raw('SELECT * FROM UNNEST (?::uuid[])', [ids]))
 
     requests.map((request) => {
       logEvent.db({


src/repositories/request-repository.ts
```diff
@@ -221,7 +225,7 @@ export class RequestRepository {
         pinned: fields.pinned,
         updatedAt: updatedAt,
       })
-      .whereIn('id', ids)
+      .where('id', 'in', connection.raw('SELECT * FROM UNNEST (?::uuid[])', [ids]))
```
stbrody, line 228: I don't really understand what this is doing well enough to comment on the pros/cons of this approach.  Are there any downsides to this type of update vs a normal update?  Why does this get around the limitation?  Could this wind up getting limited in the same way in a future release of postgres?

Just looks like of weird and brittle to me, like we're relying on an implementation detail to hack our way around how postgres wants to be used.

stbrody (commented):
would a third approach be to limit the number of *streams* in the batch to however many allow us to have less than 65k *requests?

stephhuynh18:
> would a third approach be to limit the number of _streams_ in the batch to however many allow us to have less than 65k _requests_?

thats fine with me! Just means the batch won't be full but again this would be very unlikely. The way i've implemented we could still have 1 stream overlap between batches.

src/repositories/request-repository.ts
```diff
@@ -152,7 +157,14 @@ const findRequestsToAnchorForStreams = (
   streamIds: string[],
   now: Date
 ): Promise<Array<Request>> =>
-  findRequestsToAnchor(connection, now).whereIn('streamId', streamIds).orderBy('createdAt', 'asc')
+  findRequestsToAnchor(connection, now)
+    .whereIn('streamId', streamIds)
+    .orderByRaw(
```
stbrody, line 162: why do we need this raw thing based on the streamids array?  Could we just orderBy stream_id?

src/repositories/request-repository.ts
```diff
@@ -152,7 +157,14 @@ const findRequestsToAnchorForStreams = (
...
+      `array_position(ARRAY[${streamIds.map(
+        (streamId) => `'${streamId}'`
+      )}], stream_id), created_at ASC`
+    )
+    .limit(POSTGRES_PARAMETERIZED_QUERY_LIMIT)
```
stbrody, line 167: ugh, unfortunately this means that if the postgres request limit falls in the middle of a group of requests for the same stream, you could wind up anchoring an older request instead of a newer one and cause the newer commits to wind up getting rolled back when the anchor commit supercedes them.

I think if you make the query sort first by stream_id and second by created_at in descending order, then even if half of the requests get cut off for a stream, at least you'll consider the newer requests first, and the next batch will pick up the older requests and hopefully see that they were actually already anchored implicitly

stephhuynh18: > ugh, unfortunately this means that if the postgres request limit falls in the middle of a group of requests for the same stream, you could wind up anchoring an older request instead of a newer one and cause the newer commits to wind up getting rolled back when the anchor commit supercedes them.

When a stream gets anchored, the tip of the stream is anchored, regardless of the commits in the requests. So we will not be anchoring at an older commit

When the next batch with the more recent requests gets anchored, this stream will be already anchored and will not be included in the merkle tree (unless a new commit has come in).

The problem I can see this exasperating is the zktruth error (https://linear.app/3boxlabs/issue/CDB-1960/request-status-for-non-tip-commit-throwing) as the anchor commit stored in the postgres will reference an older request. But we already have this problem for several scenarios, and when a fix is implemented this problem will go away.

We can sort the created_at in descending order BUT it may cause this scenario:
MyStream has requests for Commit1 <-- Commit2 <-- ... <-- Commit10000
The request for Commit1 is expiring. 
We create Batch1 and anchor Commit5000 <-- ... <-- Commit1000. We mark all those requests as COMPLETED
We create Batch2 immediately because Commit1 is expiring. Batch 2 is not full but had to run because the request for Commit1 is expiring.

All of this again is quite unlikely. 

And then when ceramic w/o a CAS node should get rid of this problem =)

src/repositories/request-repository.ts
stbrody: ping

src/repositories/request-repository.ts
stbrody: yeah, although anchoring the tip depends on the Ceramic node correctly having the current tip.  Having the request in the batch means the CAS will ensure that the Ceramic node has considered the request's commit if it doesn't otherwise show up in the stream log.

But this is a rare enough thing to have happen in the first place and that's another edge case on top, so I guess it doesn't really matter that much.

src/repositories/request-repository.ts
stephhuynh18: We do this because the streamIds array is sorted based on the earliest request for each streamId (ascending). Since we want to anchor expiring requests firsts we have to do this weird raw thingy. 

If we were to order by stream_id. We may not pick up expiring requests.

stbrody: can we add a comment explaining that?

stbrody (approved):

PR #872, opened 2022-12-30
feat: Add metadata entry from genesis commit on anchor submission

ukstv:
Beware: the PR is big. [This YouTube video](https://www.youtube.com/watch?v=AbSehcT19u0) is a good depiction of a development process.

## Goal

- When getting an anchor submission, CAS should store metadata information from genesis to a database.

## Constraints

- Loading genesis commit from IPFS might take a long time, so we should do it the first time we see genesis CID
- When we see genesis CID second time and more, we should update `usedAt` timestamp.

### Serialisation/deserialization aka serde

When approaching the task I was okay to reuse the same approach we use elsewhere in CAS codebase: use single Metadata entry. There is a caveat though. When creating a fresh metadata entry in the database we supply only streamId and genesis fields, gathered from genesis commit. When we retrieve it from the database and manipulate it further, it also contains timestamps. Due to how Knex+Pg+JS work with timestamps, we would need to serialize the timestamps manually to ISO string. ISO string is a ISO8601 representation of a timestamp in UTC time zone, like `2022-12-13T14:15:16.789Z`. Also, we would need to serialize and deserialize Uint8Array of `model` field, as well as StreamID instances. The serialization/deserialization could be done manually, i.e. through a hand-crafted function. The function turns to be quite complicated and opaque quite fast, especially considering that we have do serde uniformly, i.e. all dates should be converted to ISO strings same way.

Same (de)serialization complexity is applied to genesis fields. We expect genesis commit to contain certain fields. In practice we witnessed some pretty annoying cases of genesis being out of shape, like `controllers` containing random trash. If we are not strict when reading genesis fields we risk anchoring garbage.

### Sound types

Above I mentioned that there are effectively two shapes of Metadata entry. One is fresh, containing only streamID and genesis fields. Another one comes from a database. It contains timestamps. CAS is a quite complicated piece of software. To maintain sanity and avoid silly errors we better leverage type system provided by TypeScript.

Same argument applies to genesis fields. We are not interested in whatever comes from IPFS. We want particular fields of particular shape.

### Updating `usedAt` 

We expect a lot of streams to be anchored. For every stream we cache genesis fields in a database. That can not grow unboundedly. After some period we should remove the database entries considering inactive. Thus, we should have information about when it was accessed, including when it was last stored.  

We also would like to avoid having duplicate work. If we have metadata entry in the database for a StreamID, we should not re-request it from IPFS. It could be achieved naively: see if the entry is present, only store if it is absent. If an entry is present, we should update `usedAt` field. So, two SQL queries if an entry is present. There is a nice trick: `UPDATE` SQL query returns number of rows affected. Instead of doing two queries, we issue just one `UPDATE` query. If it returns `1` , it indicates the entry is stored. We reduce number of SQL queries by one.

## Building a case for additional dependencies

I had to add few dependencies to the codebase to get sound types and sane serde, namely `io-ts` , `fp-ts` and `conditional-type-checks`.

`conditional-type-checks` provides a collection of types that are quite useful for building a sound type system for an application. We are interested in `Assert` and `IsExact` types. We could just copy the corresponding code, but it seems more maintainable to use a well-supported library.

`fp-ts` is a peer dependency of `io-ts`. If we use `io-ts` we have to include `fp-ts` as well.

Now to the most interesting thing: `io-ts`. Despite an additional load of `fp-ts` it is the most well-established library in TS/JS land for declarative serde. And it is _fast_. The code it allows to write is well-read: see [models/metadata.ts file](https://github.com/ceramicnetwork/ceramic-anchor-service/compare/feature/cdb-2097-add-timestamp-if-absent...feature/cdb-2098-add-metadata-from-ipfs?expand=1#diff-59b8752a1d1722e73381bb3a414c384fe90d7749a98e681cae87ad1653432a88). The benefits in my opinion overweigh the burden of having yet another dependency.


src/ancillary/io-ts-extra.ts
```diff
@@ -0,0 +1,96 @@
...
+import * as uint8arrays from 'uint8arrays'
+import { isDIDString } from './did-string.js'
+
+/**
+ * io-ts codec for JS `Uint8Array`.
+ */
+export const uint8array = new t.Type<Uint8Array, Uint8Array, unknown>(
+  'Uint8Array',
+  (input: unknown): input is Uint8Array => input instanceof Uint8Array,
+  function (this: t.Type<Uint8Array, Uint8Array, unknown>, input: unknown, context: t.Context) {
```
pawartur, line 12: Out of curiosity: why now use an arrow function expression like in other places :) ?

ukstv: An arrow function is bound to an outside `this`. Here though I need to use a different `this` and I want to command that to the TS compiler.

pawartur (approved):
Looks good, with a couple of questions :).

One additional question: does this code still use the ceramic service for anything?

src/controllers/request-controller.ts
```diff
@@ -73,22 +83,25 @@ export class RequestController {
       }
       const streamId = StreamID.fromString(req.body.streamId)
 
+      // Store metadata from genesis to the database
+      await this.metadataService.fill(streamId)
```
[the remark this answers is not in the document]
ukstv: If not this line, CAS would not have metadata on the database at all. Now, it would not break anything. After Merkle construction is changed, absence of the line would break everything.

src/repositories/__tests__/request-repository.test.ts
```diff
@@ -81,6 +81,8 @@ describe('request repository test', () => {
       requestRepository.createOrUpdate(request),
       requestRepository.createOrUpdate(request),
     ])
+    delete result1.timestamp
```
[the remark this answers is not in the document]
ukstv: It should be same as `createdAt`, `updatedAt`. Apparently, this gave me a hint to make the tests more clear without these `delete` lines.

src/repositories/metadata-repository.ts
```diff
@@ -0,0 +1,100 @@
...
+   */
+  retrieve(streamId: StreamID): Promise<StoredMetadata | undefined>
+  /**
+   * Return true if there is a row for `streamId`.
+   */
+  isPresent(streamId: StreamID): Promise<boolean>
+  /**
+   * Mark an entry as used `now`. Return true if touched, i.e. if the entry was in the database.
+   */
+  touch(streamId: StreamID, now?: Date): Promise<boolean>
```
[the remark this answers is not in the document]
ukstv: It's just one `UPDATE` SQL query vs heavy `SELECT` _and_ `UPDATE`.

src/controllers/request-controller.ts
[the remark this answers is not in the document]
pawartur: Thanks! Do you plan to encapsulate all of the calls needed for the new Merkle tree construction to work in a single interface/method? If so, would it make sense to add a TODO here?

src/models/metadata.ts
```diff
@@ -0,0 +1,63 @@
+import * as t from 'io-ts'
+import * as te from '../ancillary/io-ts-extra.js'
+
+/**
+ * Metadata fields gathered from genesis commit.
+ */
+export const GenesisFields = t.intersection(
```
stbrody, line 7: what would happen if in the future we add fields to the genesis header for some streamtype (in js-ceramic), but forget to add them here?

ukstv: These fields will be ignored.

stbrody: will that cause any behavior to be incorrect?  Like, will the fields actually be missing from the genesis structure we write to the database and use in the CAS?  If so, could that cause any problems, like making us recreate the wrong genesis commit CID anywhere?

Or is just that we don't get type checking on those fields, but they're still passed around in the background?

src/repositories/metadata-repository.ts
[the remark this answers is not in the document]
ukstv: @pawartur Changed to `isPresent`, and commented a test for `touch` method. Hope it is now clearer.

Garbage collection is own feature.

src/models/metadata.ts
ukstv: These additional fields would be _completely_ ignored. We only use fields values in CAS, and only those fields mentioned in the code. We do not try to reconstruct CID or IPLD block from the fields for any purpose.

IMO, what should happen, is types and serde functions related to commits should be shared from `common` package.

stbrody: okay, thanks for clarifying.

stbrody:
huh, I don't seem to be able to remove myself from the reviewer list since I've commented, but consider me not a reviewer on this.

PR #885, opened 2023-01-13
fix: only instantiate contract if contract address given

stephhuynh18:


v-stickykeys (commented):
The error handling here seems a bit redundant and hard to follow.

Can we simplify to just

If usesmartcontractanchors { if contractaddress {...} else { throw } }

stephhuynh18:
> The error handling here seems a bit redundant and hard to follow.
> 
> Can we simplify to just
> 
> If usesmartcontractanchors { if contractaddress {...} else { throw } }

@v-stickykeys yes it is =( I was being lazy HAHA because of the tests dislikes the option you provided. But i should do it right.

gvelez17:
hey @stephhuynh18 should we fix conflicts and merge this during our on-call week?  figured cleanup old branches

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -151,10 +151,22 @@ export class EthereumBlockchainService implements BlockchainService {
...
+    if (
+      !this.config.blockchain.connectors.ethereum.contractAddress &&
+      this.config.useSmartContractAnchors
+    ) {
+      throw Error(
+        'You must set a smart contract address if you would like to use smart contract anchors'
+      )
+    }
+
+    if (this.config.blockchain.connectors.ethereum.contractAddress) {
```
gvelez17, line 164: we don't need this if since we have the throw above?  but ya i like the clear error message

PR #886, opened 2023-01-13
fix: array_position needs explicit varchar type

stephhuynh18:
odd tests didn't catch this


stbrody (approved):
huh, I also wonder why tests didn't catch it, that's a little concerning

PR #889, opened 2023-01-17
fix: make block info in anchor proof configurable

smrz2001:
I haven't added config options to CAS before so not fully sure if I've covered all the changes. Please let me know if I've missed something.

Also, is this something I should add a unit test for? `anchor-service.test.ts` seemed like the right place, if so.

### DIFF
--- a/config/env/dev.json
+++ b/config/env/dev.json
@@ -5,6 +5,7 @@
   "anchorControllerEnabled": "@@ANCHOR_CONTROLLER_ENABLED",
   "anchorLauncherUrl": "@@ANCHOR_LAUNCHER_URL",
   "expirationPeriod": "@@ANCHOR_EXPIRATION_PERIOD",
+  "includeBlockInfoInAnchorProof": "@@INCLUDE_BLOCK_INFO_IN_ANCHOR_PROOF",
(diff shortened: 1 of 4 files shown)


config/env/dev.json
```diff
@@ -5,6 +5,7 @@
   "anchorControllerEnabled": "@@ANCHOR_CONTROLLER_ENABLED",
   "anchorLauncherUrl": "@@ANCHOR_LAUNCHER_URL",
   "expirationPeriod": "@@ANCHOR_EXPIRATION_PERIOD",
+  "includeBlockInfoInAnchorProof": "@@INCLUDE_BLOCK_INFO_IN_ANCHOR_PROOF",
```
stbrody, line 8: is there a clay.json we need to update also?

stbrody (approved):
lgtm

config/env/dev.json
smrz2001: Good question, there are 3 files: `dev.json`, `prod.json`, and `test.json`. No separate file for Clay.

smrz2001: Just also checked that previously added config values went into `dev.json` and `prod.json` and not in other places.

PR #894, opened 2023-01-18
chore: Beautify IpfsService

ukstv:
Added:
- `IpfsService` as an implementation from `IIpfsService` as an interface live in different files
- `IpfsService` accepts `AbortSignal`
- `IpfsService` passes native IPFS-client timeouts
- `IpfsService::retrieveRecord` accepts `path` option to use IPLD traversal
- `Utils.delay` accepts `AbortSignal` too
- `IpfsService` has a test suite

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -3,21 +3,32 @@ import { create as createMultihash } from 'multiformats/hashes/digest'
 
 const KECCAK_256_CODE = 0x1b
 const ETH_TX_CODE = 0x93
+
+/**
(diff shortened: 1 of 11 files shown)


src/utils.ts
```diff
@@ -36,10 +47,9 @@ export class Utils {
    * @param arr - Array of number
    */
   static averageArray(arr: number[]): number {
-
-    if (arr.length == 0) {
-        return 0
-    }    
+    if (arr.length === 0) {
```
ukstv, line 50: Prettier.

stephhuynh18 (approved):

PR #899, opened 2023-01-19
fix: apply loadStream timeout to the multiquery request also

stbrody:
is this actually safe?  The multiqueries are expected to take longer than the regular loadStream since they can never just get the state from the state store, they always have to actually apply the log for the commits manually.  So this does seem like maybe we could get stuck with the multiquery always timing out and timing out again on retry.

my brain is too fried to think through this fully at the moment, but if we can convince ourselves this is safe, it could help speed up anchor batches even more


ukstv (approved):

stbrody:
Closing this in favor of https://github.com/ceramicnetwork/ceramic-anchor-service/pull/901 which seems like a safer approach

PR #901, opened 2023-01-20
fix: Optimize multiquery to only load newest missing request

stbrody:
This hasn't been tested at all, please review carefully in case I missed anything

### DIFF
--- a/src/services/anchor-service.ts
+++ b/src/services/anchor-service.ts
@@ -772,9 +772,14 @@ export class AnchorService {
       stream = await ceramicService.loadStream(candidate.streamId)
     } catch (err) {
       logger.err(`Failed to load stream ${candidate.streamId.toString()}: ${err}`)
+      Metrics.count(METRIC_NAMES.FAILED_STREAM, 1)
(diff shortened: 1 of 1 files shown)


ukstv (approved):

src/services/anchor-service.ts
```diff
@@ -772,65 +772,95 @@ export class AnchorService {
...
+    const stillMissingRequests = candidate.requests.filter((req) => {
+      const found = stream.state.log.find(({ cid }) => {
+        return cid.toString() == req.cid
+      })
+      return !found
+    })
+
     // Fail requests for tips that failed to be loaded
-    for (const request of missingRequests) {
+    for (const request of stillMissingRequests) {
```
stephhuynh18, line 857: I believe these requests will be failed and failed due to conflict resolution (rejected). Is that the desired behavior

stbrody: yes that's the behavior.  Does the failure reason matter at this point?  I forget if we still do something based on the failure reason message

PR #903, opened 2023-01-20
DID based authentication using API gateway

v-stickykeys:
## Description

This PR adds a Serverless application that deploys AWS Lambda functions to an API Gateway which proxies all requests from `cas-<env>.3boxlabs.com` to the internal load balancer that sits in front of the CAS API containers.

## Motivation

This way of handling requests to CAS keeps a separation between the CAS API logic further and the logic for batching and anchoring. With this approach we take advantage of AWS API Gateway infrastructure/tooling to support authentication and rate-limiting, making it easy to modify these without needing to redeploy CAS.

## Changes

- Adds env var `REQUIRE_AUTH` to CAS
- When above var is true, verifies any request body hashes to the `digest` in the header of the request (note: this header is provided by api gateway, not the ceramic daemon)


auth/src/authorizers/default.ts
```diff
@@ -0,0 +1,96 @@
...
+    const ip = event.requestContext.identity.sourceIp
+    console.log('ip', ip)
+    const request = ec2.describeSecurityGroups({
+      Filters: [{Name: 'ip-permission.cidr', Values: [`${ip}/32`] }],
+      GroupIds: permissionedSecurityGroupIds
+    })
+    try {
+      const data = await request.promise()
+      console.log(data)
+      if (!data.SecurityGroups) throw new Error('No security groups found with this IP')
```
gvelez17, line 53: should we throw an error or just return unauthorized?

auth/src/authorizers/default.ts
```diff
@@ -0,0 +1,96 @@
...
+      if (data.SecurityGroups.length < 1) throw new Error('No security groups found with this IP')
+      return callback(null, generatePolicy(ip, {effect: 'Allow', resource: event.methodArn}, ip))
+    } catch (err) {
```
gvelez17, line 56: should we check here if the error is not found?

gvelez17 (approved):
Approved- additional changes can be addressed in follow-up prs

oed:
This PR doesn't enable the DID Auth checks yet right?

auth/src/services/aws/dynamodb.ts
```diff
@@ -0,0 +1,553 @@
...
+                { AttributeName: 'PK', KeyType: 'HASH' },
+                { AttributeName: 'SK', KeyType: 'RANGE' }
+            ],
+            AttributeDefinitions: [
+                { AttributeName: 'PK', AttributeType: 'S' },
+                { AttributeName: 'SK', AttributeType: 'S' },
+                { AttributeName: 'GSI-1-PK', AttributeType: 'S' },
+                { AttributeName: 'GSI-1-SK', AttributeType: 'S' },
+            ],
+            BillingMode: BillingMode.PAY_PER_REQUEST,
```
smrz2001, line 101: Is this the optimal billing mode?

auth/src/services/aws/dynamodb.ts
```diff
@@ -0,0 +1,553 @@
...
+        const input: CreateTableCommandInput = {
+            TableName: tableName,
+            KeySchema: [
+                { AttributeName: 'PK', KeyType: 'HASH' },
+                { AttributeName: 'SK', KeyType: 'RANGE' }
+            ],
+            AttributeDefinitions: [
+                { AttributeName: 'PK', AttributeType: 'S' },
+                { AttributeName: 'SK', AttributeType: 'S' },
+                { AttributeName: 'GSI-1-PK', AttributeType: 'S' },
```
smrz2001, line 98: I don't see the GSIs being used here, do we need them?

gvelez17:
> 

I believe it does but it also allows IP login, is my understanding

auth/src/authorizers/default.ts
v-stickykeys: Here we throw an error which is caught below so that in the lambda logs it is clear why the call returns as "unauthorized"

auth/src/authorizers/default.ts
v-stickykeys: I don't believe there is a "not found error" for the security groups, so just checking we get any security groups matching the description.

If an unknown error is returned it gets handled and logged below.

auth/src/services/aws/dynamodb.ts
v-stickykeys: They are added by default by serverless lift, so added the same schema here for consistency if we create a table outside of lift. Any harm in keeping them even if not used?

auth/src/services/aws/dynamodb.ts
v-stickykeys: It seems straight forward to start. I think we could do provisioned but not sure how much would be appropriate. Suggestions are welcome!

v-stickykeys:
> This PR doesn't enable the DID Auth checks yet right?

This PR now includes checks that the request body has not be altered.
It also includes checking DIDs are registered OR that the IP address is already on our allowlist.

ukstv (approved):

src/auth/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+function buildBodyDigest(contentType: string, body: any): string | undefined {
+    if (!body) return
+
+    let hash: Hash
+
+    if (contentType) {
+      if (contentType.includes('application/vnd.ipld.car')) {
+        const carFactory = new CARFactory()
+        const car = carFactory.fromBytes(u8a.fromString(body, 'binary'))
```
v-stickykeys, line 27:
```suggestion
        const car = carFactory.fromBytes(body)
```

auth/src/services/aws/dynamodb.ts
smrz2001: I think that should be ok. In this case, since we're never writing to any fields that are part of the GSI, we probably won't be charged for their throughput.

When you get a chance, would you mind describing how the tables/indices are used (i.e. the read/write patterns)? Would to good to do a quick analysis to make sure there aren't any hot spots.

v-stickykeys: Yes! will add to your calendar to discuss

src/auth/index.ts
```diff
@@ -0,0 +1,54 @@
...
+        hash = new sha256.SHA256().update(u8a.fromString(JSON.stringify(body)))
+      }
+    }
+
+    if (!hash) {
+      // Default to hashing stringified body
+      hash = new sha256.SHA256().update(u8a.fromString(JSON.stringify(body)))
+    }
+
+    return `0x${u8a.toString(hash.digest(), 'base16')}`
```
v-stickykeys, line 53: @ukstv mind taking a look over this file which has changed?

src/auth/index.ts
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+        next()
+    }
+}
+
+function buildBodyDigest(contentType: string, body: any): string | undefined {
+    if (!body) return
+
+    let hash: sha256.SHA256
```
ukstv, line 36:
```suggestion
    let hash: Uint8Array
```

src/auth/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    if (contentType) {
+      if (contentType.includes('application/vnd.ipld.car')) {
+        const carFactory = new CARFactory()
+        const car = carFactory.fromBytes(body)
+        return car.roots[0].toString()
+      } else if (contentType.includes('application/json')) {
+        hash = new sha256.SHA256().update(u8a.fromString(JSON.stringify(body)))
```
ukstv, line 44:
```suggestion
        hash = sha256.hash(u8a.fromString(JSON.stringify(body)))
```

src/auth/index.ts
```diff
@@ -0,0 +1,54 @@
...
+      }
+    }
+
+    if (!hash) {
+      // Default to hashing stringified body
+      hash = new sha256.SHA256().update(u8a.fromString(JSON.stringify(body)))
```
ukstv, line 50:
```suggestion
      hash = sha256.hash(u8a.fromString(JSON.stringify(body)))
```

src/auth/index.ts
ukstv, line 53:
```suggestion
    return `0x${u8a.toString(hash, 'base16')}`
```

ukstv (changes requested):

PR #909, opened 2023-01-23
fix: use loadStream with SYNC_ON_ERROR flag for pinning

stbrody:
This will help push the time spent resycing streams with CACAO errors to happen out-of-band when requests come in, rather than during the anchor batches as they run.


ukstv (approved):

ukstv:
wouldn't hurt

PR #919, opened 2023-01-24
 feat: don't do the multiquery to force loading missing commits at all

stbrody:
THIS HAS NOT BEEN TESTED. Definitely test this before deploying to prod!


ukstv (approved):
Hope it really works as intended :)

PR #927, opened 2023-01-30
fix: don't pin when loading stream for anchor worker

stbrody:
We'll still try to pin the stream during request creation in the API server


smrz2001 (approved):
:shipit:

PR #941, opened 2023-02-09
Trim docker image size

dbcfd:
Adjust base image, use layers to further reduce size.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,4 @@ node_modules
 .env
 .git
 tmp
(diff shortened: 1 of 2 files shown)


gvelez17:
hey thanks for sharing this with me @dbcfd !  I am applying these ideas.  Should we turn it back into a draft for now on cas or do we need to resolve this thing

```
8:49PM FATAL system | failed to execute plan: task failed: actions._baseImage._build: target stage base could not be found
```
?

PR #945, opened 2023-02-15
chore: Use same strict tsconfig as in js-ceramic anchor-utils

ukstv:


src/merkle/merkle-tree.ts
```diff
@@ -62,7 +63,7 @@ export class MerkleTree<TData, TLeaf extends TData, TMetadata> {
     /**
      * Tree metadata
      */
-    readonly metadata: TMetadata
+    readonly metadata: TMetadata | null
```
pawartur, line 66: Can you add a comment about what it means that metadata is null?

src/normalize-config.util.ts
```diff
@@ -29,7 +29,9 @@ export function normalizeConfig(config: Config): void {
  */
 export function cleanupConfigForLogging(config: Config): Record<string, any> {
   const configCopy = cloneDeep(config)
```
pawartur, line 31: Just a thought, not really a comment to this PR: seeing that we have a function called `cleanupConfigForLogging(...)`, I'd be seriously worried that somebody in future sees that we log configs and logs it themselves, but forgets to put it through the function first.

I guess, I'd prefer to have a custom logging function that we'd use everywhere which would make sure that all the sensitive data is hidden, before putting in logs.

pawartur (approved):

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -66,20 +66,15 @@ async function attempt<T>(
  * @param walletBalance - Available funds.
  */
 function handleInsufficientFundsError(txData: TransactionRequest, walletBalance: BigNumber): void {
-  const txCost = (txData.gasLimit as BigNumber).mul(txData.maxFeePerGas)
+  const txCost = (txData.gasLimit as BigNumber).mul(txData.maxFeePerGas || 1)
```
pawartur, line 69: Can you put this `1` as a constant somewhere? Out of curiosity, why `1` is the right number :) ?

ukstv: That's to tell TS it's a safe operation. Going to add a comment there.

ukstv: Or, rather change the code.

src/merkle/merkle-tree.ts
ukstv: Added: "Tree metadata. Would not be considered for a proof if not available."

src/normalize-config.util.ts
ukstv: I'd set the probability of anyone misusing logging for this to be close to zero, as it requires pretty deep penetration inside the codebase.

PR #954, opened 2023-02-24
CAS Auth Fixes

v-stickykeys:
### Summary
I found a few issues with CAS auth in dev. After applying these fixes to a [temporary image](https://github.com/ceramicnetwork/ceramic-anchor-service/tree/temp/debug) I did not see any unrecognizable errors.

### Changes
- Get use `digest` property from signed payload instead of body (to match [how js-ceramic creates payload](https://github.com/ceramicnetwork/js-ceramic/blob/eb179830da3b5396794e85a5df374c7fc1590829/packages/core/src/anchor/auth/did-anchor-service-auth.ts#L49-L53))
- Check if `req.body` is an empty object in auth middleware (skip verification if so)
- Add return statements so `next` is not called twice in auth middleware

### DIFF
--- a/auth/src/authorizers/default.ts
+++ b/auth/src/authorizers/default.ts
@@ -81,7 +81,7 @@ async function allowRegisteredDID(event: APIGatewayRequestAuthorizerEvent, callb
   if (result) {
     const did = result.didResolutionResult.didDocument?.id
     const nonce = result.payload?.nonce
-    const digest = result.payload?.body
+    const digest = result.payload?.digest
     if (!did) {
       console.error('Missing did')
     } else if (!nonce) {
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -18,11 +18,13 @@ function buildExpressMiddleware() {
     return function(req: Request, _res: Response, next: NextFunction) {
         if (req.headers) {
             if (req.headers['did'] && req.body) {
-                const digest = buildBodyDigest(req.headers['content-type'], req.body)
-                if (req.headers['digest'] == digest) {
-                  next()
-                } else {
-                  throw Error('Body digest verification failed')
+                if (Object.keys(req.body).length > 0) {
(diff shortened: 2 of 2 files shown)


oed (approved):

ukstv (approved):

ukstv:
@pawartur Mind testing with @v-stickykeys if CAR file goes through after this is merged, please?

auth/src/authorizers/default.ts
```diff
@@ -81,7 +81,7 @@ async function allowRegisteredDID(event: APIGatewayRequestAuthorizerEvent, callb
   if (result) {
     const did = result.didResolutionResult.didDocument?.id
     const nonce = result.payload?.nonce
-    const digest = result.payload?.body
+    const digest = result.payload?.digest
```
nathanielc, line 84: I think we still my have a bug, this code adds the payload but I am not seeing the `digest` in the JWS payload when I decode it manually. 

https://github.com/ceramicnetwork/js-ceramic/blob/eb179830da3b5396794e85a5df374c7fc1590829/packages/core/src/anchor/auth/did-anchor-service-auth.ts#L49-L53

Do you have any ideas on why the JWS payload coming from the Ceramic node does not have the `digest`?

v-stickykeys: The authorizer lambda is responsible for passing this digest in a context object [here](https://github.com/ceramicnetwork/ceramic-anchor-service/blob/develop/auth/src/authorizers/default.ts#L99-L102). Then API gateway adds the header [here](https://github.com/3box/ceramic-infra/blob/dev/app-cas-auth/modules/did-auth/api_gateway.tf#L39-L41).

v-stickykeys: What do you mean "when I decode it manually"? What are you decoding and how?

The payload from ceramic may not have a `digest` property if the request does not include a `body`.

PR #955, opened 2023-02-25
Rate limit access to IPFS to avoid overloading it

stbrody:
Right now if we get a spike of concurrent request to the CAS API, it can overwhelm the IPFS node, causing the node to have a huge memory spike and crash (see grafana screenshots below).  Instead of sending more requests to IPFS than it can handle, we can slow down the processing of incoming requests to avoid overloading the IPFS node.

Screen Shot 2023-02-25 at 1 17 49 PM
Screen Shot 2023-02-25 at 1 17 59 PM

### DIFF
--- a/src/services/ipfs-service.ts
+++ b/src/services/ipfs-service.ts
@@ -13,6 +13,7 @@ import * as https from 'https'
 import { PubsubMessage } from '@ceramicnetwork/core'
 import type { IIpfsService, RetrieveRecordOptions } from './ipfs-service.type.js'
 import type { AbortOptions } from './abort-options.type.js'
+import { Semaphore } from 'await-semaphore'
 
 const { serialize, MsgType } = PubsubMessage
 
@@ -49,6 +50,7 @@ export class IpfsService implements IIpfsService {
   private readonly cache: LRUCache<string, any>
   private readonly pubsubTopic: string
   private readonly ipfs: IPFS
+  private readonly semaphore: Semaphore
 
   static inject = ['config'] as const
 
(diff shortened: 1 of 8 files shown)


src/services/ipfs-service.ts
```diff
@@ -110,21 +115,18 @@ export class IpfsService implements IIpfsService {
...
-        record,
-        { signal: options.signal, timeout: IPFS_PUT_TIMEOUT }
-    ).then(
-        // Note: While dag.put has a pin flag it always recurses and
-        // we do not want to recurse so we explicitly call pin.add.
-        (cid) => this.ipfs.pin.add(cid, {
-            signal: options.signal,
-            timeout: IPFS_PUT_TIMEOUT,
-            recursive: false
+    return this.ipfs.dag.put(record, { signal: options.signal, timeout: IPFS_PUT_TIMEOUT }).then(
```
stbrody, line 118: this is just prettier cleanup noise

stbrody:
@ukstv I'm not sure what's causing the compile failure in await-semaphore or what to do about it, esp since it's using the same version of the package that we use successfully in js-ceramic.

src/services/ipfs-service.ts
dbcfd, line 118: Do we need to rate limit writes?

stbrody: each anchor worker runs in its own process, so there's no shared memory so a semaphore won't help.  Each individual worker does writes to ipfs entirely serially, the only parallel access comes from running multiple workers in parallel.  But we only run 2 workers currently, maybe we'll bump to 3 or 4 at some point, but I'm still not too worried about the writes overloading the ipfs node.  The reads cause far more simultaneous load as there's one read for every incoming anchor request, and we get hundreds of requests a second.

dbcfd (approved):

PR #968, opened 2023-03-01
fix(auth): move ip allow list check ahead of auth header verification

smrz2001:


auth/src/authorizers/default.ts
```diff
@@ -54,8 +57,9 @@ async function allowPermissionedIPAddress(event: APIGatewayRequestAuthorizerEven
     if (ALLOWED_IP_ADDRESSES[ip]) {
       return callback(null, generatePolicy(ip, {effect: 'Allow', resource: event.methodArn}, ip))
```
smrz2001, line 58: Can this return null?!

v-stickykeys (approved):
LGTM

PR #991, opened 2023-03-15
chore: Expose REPLACED on CAS side

ukstv:


src/services/request-presentation-service.ts
```diff
@@ -98,7 +98,6 @@ export class RequestPresentationService {
       case RequestStatus.PROCESSING:
       case RequestStatus.FAILED:
       case RequestStatus.READY:
-        return this.notCompleted(request, status)
```
stbrody, line 101: I'm confused by this change.  Wouldn't this make all of PENDING, PROCESSING, READY, and REPLACED all come back to Ceramic as FAILED?  I would have thought that the change to make would be removing the logic that overrides the `status` field on line 106

ukstv: Agree. Rebase went wrong.

stbrody: seems like tests should have caught that?  Can we add some test coverage that would have prevented us from accidentally putting in a regression like that if I hadn't noticed the issue in code review?

ukstv: They did catch that :) ceramic integration test failed because of that.

ukstv:
@stbrody all green now.

stbrody (approved):
lgtm

PR #1004, opened 2023-04-04
fix(deps): CDB-2128 Address nonce errors

dbcfd:
Per https://github.com/ethers-io/ethers.js/issues/3812, nonce may be incorrectly incremented, which can produce off by one errors which we are currently seeing in QA and Clay


src/auth/index.ts
```diff
@@ -16,50 +16,53 @@ function buildExpressMiddleware() {
...
-                if (Object.keys(req.body).length > 0) {
-                    const digest = buildBodyDigest(req.headers['content-type'], req.body)
-                    if (req.headers['digest'] == digest) {
-                      return next()
-                    } else {
-                      throw Error('Body digest verification failed')
-                    }
-                }
-            }
+  return function (req: Request, _res: Response, next: NextFunction) {
```
dbcfd, line 19: Prettier

src/settings.ts
```diff
@@ -39,7 +38,7 @@ export enum METRIC_NAMES {
   // Anchor Service (histograms)
 
   // request that moves from ready -> processing
-  READY_PROCESSING_MS = "ready_processing_ms",
+  READY_PROCESSING_MS = 'ready_processing_ms',
```
dbcfd, line 41: Prettier

src/models/transaction.ts
```diff
@@ -2,12 +2,12 @@ export class Transaction {
   chain: string
   txHash: string
   blockNumber: number
-  blockTimestamp: number
+  blockDate: string
```
stbrody, line 5: if we're getting a Date object back from ethers can we keep using a Date type throughout our code?  That's a stronger type match than just using string.

src/services/anchor-service.ts
```diff
@@ -404,7 +404,7 @@ export class AnchorService {
     if (this.includeBlockInfoInAnchorProof) {
       ipfsAnchorProof = {
         blockNumber: tx.blockNumber,
-        blockTimestamp: tx.blockTimestamp,
+        blockDate: tx.blockDate,
```
stbrody, line 407: changing the fieldname and the type from a timestamp to an ISO string would change the format of the anchor proof, which could break Ceramic as it won't know how to interpret the new anchor proof format.  Can we keep this as it was?

src/utils.ts
```diff
@@ -41,4 +41,28 @@ export class Utils {
     const cidVersion = 1
     return CID.create(cidVersion, ETH_TX_CODE, multihash)
   }
+
+  /**
+   * Coalesce a possible null value to a value or error
+   * @param func
+   * @param error
+   */
+  static coalesce<T>(func: () => T | null | undefined, error: string): T {
```
stbrody, line 50: `checkNotNull`?  it wasn't super obvious to me from reading the calling code what "coalesce" meant

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -119,16 +129,15 @@ function make(config: Config): EthereumBlockchainService {
   let provider
   if (url) {
     logger.imp(`Connecting ethereum provider to url: ${url}`)
-    provider = new ethers.providers.StaticJsonRpcProvider(url)
+    provider = new ethers.JsonRpcProvider(url)
```
stbrody, line 132: the reason we were using `StaticJsonRpcProvider` was to avoid excessive numbers of calls to `eth_chainid` as per this bug:  https://github.com/ethers-io/ethers.js/issues/901

If the static version doesn't exist anymore, we should make sure to double-check that this change doesn't result in a ton of eth_chainid calls again.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -201,48 +219,47 @@ export class EthereumBlockchainService implements BlockchainService {
...
       const nextPriorityFee = EthereumBlockchainService.increaseGasPricePerAttempt(
         maxPriorityFeePerGas,
-        attempt,
         prevPriorityFee
       )
       txData.maxPriorityFeePerGas = nextPriorityFee
-      const baseFee = maxFeePerGas.sub(maxPriorityFeePerGas)
-      txData.maxFeePerGas = baseFee.add(nextPriorityFee)
+      // must be greater than baseFee + maxPriorityFeePerGas
+      txData.maxFeePerGas = addPadding(baseFee + nextPriorityFee)
```
stbrody, line 244: why are we adding padding here if we already added padding in `increaseGasPricePerAttempt`?  Is this redundant?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -260,23 +277,13 @@ export class EthereumBlockchainService implements BlockchainService {
...
-  ): BigNumber {
-    // Try to increase an estimated gas price first
-    const estimateBN = BigNumber.from(estimate)
-    const increase = estimateBN.div(10).mul(attempt) // 10% increase per attempt
-    const increaseEstimate = estimateBN.add(increase)
-
-    if (attempt == 0 || previousGas == undefined) {
-      return increaseEstimate
+  ): bigint {
+    if (previousGas == undefined) {
```
stbrody, line 282: so this is a behavior change to the gas logic.  Previously we would add an increasing amount of padding per attempt on top of the gas estimate we get from ethers.  So the first attempt we'd just use the gas estimate from ethers directly.  The second attempt we'd add 10% on top of the estimate, and the third time we'd add 20% on top of the estimate.
In this new version we're just always adding 10% on top of the estimate every time.

I don't feel strongly about which logic is better, but could we possibly separate the changes to our estimation logic from the code changes required to keep the same logic with upgrading to the new version of the ethers package, each into their own PR?

stbrody (changes requested):
Most important note is to avoid changing the format of the AnchorProof.

src/models/transaction.ts
dbcfd: Sadly, the cbor encoding doesn't support it. Based on your other comments, I think the best thing is just to use number and millis since epoch.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
dbcfd: Will have to check. The static provider doesn't seem to exist anymore.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
dbcfd: On upgrading ethers, we run into underpriced transaction errors. I'm not sure if ethers changed something in the data they report back, but this is required to not have underpriced transaction errors.

src/models/transaction.ts
stbrody: I meant hold on to a Date reference in our code, then serialize it to a number before writing it out to IPFS

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
ukstv: @dbcfd There is an option that emulates `StaticJsonRpcProvider`: https://docs.ethers.org/v6/api/providers/jsonrpc/#JsonRpcApiProviderOptions

pnpm-lock.yaml
ukstv, line 1: NPM `package-lock.json` or PNPM `ppm-lock.yaml`? If PNPM, then slightly larger changes with regards to deployment are needed.

src/models/transaction.ts
dbcfd: Good suggestion, will look into changing it there.

src/services/blockchain/__tests__/eth-bc-service.test.ts
```diff
@@ -66,8 +64,12 @@ describe('ETH service connected to ganache', () => {
 
   describe('v0', () => {
     test('should send CID to local ganache server', async () => {
-      const block = await providerForGanache.getBlock(await providerForGanache.getBlockNumber())
-      const startTimestamp = block.timestamp
+      const ganacheBlockNumber = await providerForGanache.getBlockNumber()
+      const block = await Utils.coalesceAsync(
```
ukstv, line 68: There is `expectPresent` function that could be used instead.
```typescript
const block = ...
expectPresent(block);
// TS types and runtime know `block` is defined
```

src/services/blockchain/__tests__/eth-bc-service.test.ts
```diff
@@ -93,35 +91,27 @@ describe('ETH service connected to ganache', () => {
 
     test('gas price increase math', () => {
       const gasEstimate = {
-        maxFeePerGas: BigNumber.from(2000),
-        maxPriorityFeePerGas: BigNumber.from(1000),
-        gasPrice: BigNumber.from(0),
+        maxFeePerGas: BigInt(2000),
```
ukstv, line 94:
```suggestion
        maxFeePerGas: 2000n,
```

`123123n` is a bigint literal in JS.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -65,16 +71,20 @@ async function attempt<T>(
  * @param txData - Transaction to write.
  * @param walletBalance - Available funds.
  */
-function handleInsufficientFundsError(txData: TransactionRequest, walletBalance: BigNumber): void {
-  const txCost = (txData.gasLimit as BigNumber).mul(txData.maxFeePerGas!)
-  if (txCost.gt(walletBalance)) {
+function handleInsufficientFundsError(txData: TransactionRequest, walletBalance: bigint): void {
+  const gasLimit = Utils.coalesce(() => txData.gasLimit, 'No gas limit')
```
ukstv, line 75: And here simpler `if (!gasLimit) throw new Error()` seems more appropriate than coalesce.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -84,7 +94,7 @@ function handleInsufficientFundsError(txData: TransactionRequest, walletBalance:
  * Represent chainId in CAIP format.
  * @param chainId - Numeric chain id.
  */
-function caipChainId(chainId: number) {
+function caipChainId(chainId: bigint | number) {
```
ukstv, line 97: chainId is now bigint in ethers. So lovely!

ukstv (changes requested):
Seems like calculation logic is okay. Minor technical improvements could be handy.

And emulation of `StaticJsonRpcProvider`! That could require some additional refactoring though: I am not sure if `staticNetwork` would work without giving explicit network to the provider.

pnpm-lock.yaml
ukstv: If NPM, then PNPM lock file should be removed from the PR to eliminate confusion.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
dbcfd: @ukstv Looking at the docs, that looks to default to null, which would prevent the eth_chainid calls, so I don't think any change is needed here?

ukstv: @dbcfd Last time I read ethers 6 code `null` there meant that it would try to get chain id on every call. If the parameter is absent (default: `null`) means we would struggle with excessive chainId calls.

ukstv (approved):

ukstv:
@dbcfd Oh, lovely ethers: https://github.com/ceramicnetwork/ceramic-anchor-service/actions/runs/4670215624/jobs/8269672491#step:10:3555

At one point I thought maybe we could ditch ethers with [viem](https://viem.sh) altogether.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
ukstv: Yeah, I see the point.

jxom:
> At one point I thought maybe we could ditch ethers with [viem](https://viem.sh/) altogether.

viem author here! We are happy to help 😉

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -115,20 +126,50 @@ function handleTimeoutError(transactionTimeoutSecs: number): void {
 function make(config: Config): EthereumBlockchainService {
   const ethereum = config.blockchain.connectors.ethereum
   const { host, port, url } = ethereum.rpc
+  let network
+  try {
+    network = Network.from(ethereum.network)
```
stbrody, line 131: hmm... this is likely to be a breaking change as now the network option is required in the config even when using an infura url (which implies the network and so network wasn't previously required).

That's fine since we're the only ones running a CAS, just something to look out for when we go to deploy.  We should double check we have the network set so things don't break on the upgrade

stbrody: Edit: I see it's not actually required as if its not set we fall back to the non-static behavior.  But that could cause our infura endpoint to be pummeled with chainid calls and blow up our bill (this happened in the past) so we'll still want to be really careful here and make sure have the network flag set on all 3 envs, and monitor our infura account as this gets deployed to various envs.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -115,20 +126,50 @@ function handleTimeoutError(transactionTimeoutSecs: number): void {
...
+    const hostPort = `${host}:${port}`
+    if (network) {
+      provider = new ethers.JsonRpcProvider(hostPort, network, {
+        staticNetwork: network,
+      })
+    } else {
+      provider = new ethers.JsonRpcProvider(hostPort)
+    }
+  } else if (network) {
+    logger.imp(`Connecting ethereum to etherscan provider for network ${ethereum.network}`)
```
stbrody, line 157: what do you mean by "etherscan" provider here?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -115,20 +126,50 @@ function handleTimeoutError(transactionTimeoutSecs: number): void {
...
+    const opts = {
+      alchemy: "-",
+      ankr: "-",
+      cloudflare: "-",
+      etherscan: ethereum.account.privateKey,
```
stbrody, line 162: what does this do?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -115,20 +126,50 @@ function handleTimeoutError(transactionTimeoutSecs: number): void {
...
+        staticNetwork: network,
+      })
+    } else {
+      provider = new ethers.JsonRpcProvider(url)
+    }
   } else if (host && port) {
     logger.imp(`Connecting ethereum provider to host: ${host} and port ${port}`)
-    provider = new ethers.providers.StaticJsonRpcProvider(`${host}:${port}`)
+    const hostPort = `${host}:${port}`
+    if (network) {
```
stbrody, line 149: maybe we can reduce code duplication by creating an empty `options` object and adding the `staticNetwork` field to it if `network` is set.  Then we can always just pass the options object to the `JSONRpcProvider` constructor and not need the extra `if(network)` conditional in both blocks?

stbrody (changes requested):
only actually important comment is making sure that we're serializing the date timestamp into the ipfs proof correctly

src/services/anchor-service.ts
```diff
@@ -404,7 +404,7 @@ export class AnchorService {
     if (this.includeBlockInfoInAnchorProof) {
       ipfsAnchorProof = {
         blockNumber: tx.blockNumber,
-        blockTimestamp: tx.blockTimestamp,
+        blockTimestamp: tx.blockTimestamp.getUTCDate(),
```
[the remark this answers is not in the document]
dbcfd: Good catch. Will fix that and check the test.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
dbcfd: The options make it so only etherscan works, since that's we use for our tests. All other providers would need host/port, and preferably name.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
dbcfd: Removes all providers other than etherscan, and sets the private key for etherscan.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
dbcfd: Will reduce the duplication.

PR #1008, opened 2023-04-10
chore: Use codeco instead of io-ts+fp-ts

ukstv:
Replace io-ts + fp-ts with new kid on the block of coders/decoders [codeco](https://github.com/ceramicnetwork/codeco).

- Same raw speed.
- 2x-7x less bundle size. Suitable for browser code.
- More convenient optional fields:
  -  `t.sparse({a: t.optional(t.string)})` for `{a?: string}`,
  - `t.type({a: t.replacement(t.string, "hello")})` for when you want to _parse_ `hello` as a default value,
  - `t.type({a: t.defaults(t.string, "hello")})` for when you want to _use_ `hello` as a default value.

@PaulLeCam mentioning you here as a reviewer to compare codeco and io-ts.


src/ancillary/__tests__/anchor-request-params-parser.test.ts
```diff
@@ -11,8 +11,7 @@ import { bases } from 'multiformats/basics'
 import { GenesisFields } from '../../models/metadata.js'
 import { asDIDString } from '../did-string.js'
 import { mockRequest } from '../../controllers/__tests__/mock-request.util.js'
-import * as t from 'io-ts'
-import { isRight, Right } from 'fp-ts/lib/Either.js'
+import * as t from 'codeco'
```
dbcfd, line 14: Although t appeared some places before, it's used a lot more now in tests. Can we name this something else, so it's a little easier to grok when its used in tests?

src/ancillary/__tests__/io-ts-extra.test.ts
```diff
@@ -1,20 +1,20 @@
 import { describe, test, expect } from '@jest/globals'
-import * as te from '../io-ts-extra.js'
-import { isRight, Right } from 'fp-ts/lib/Either.js'
+import * as te from '../codecs.js'
+import * as t from 'codeco'
```
dbcfd, line 3: Same comment as before

src/ancillary/codecs.ts
```diff
@@ -1,39 +1,35 @@
-import * as t from 'io-ts'
 import { CID } from 'multiformats/cid'
 import { CommitID, StreamID } from '@ceramicnetwork/streamid'
 import * as uint8arrays from 'uint8arrays'
 import { isDIDString } from './did-string.js'
+import * as t from 'codeco'
```
dbcfd, line 5: Rather than a `*` import, can we just bring in `TrivialCodec` and `IContext`? If we do need a `*` import, rename to something like `cdc` or `codec`.

src/controllers/request-controller.ts
```diff
@@ -13,10 +13,8 @@ import {
   RequestAnchorParams,
 } from '../ancillary/anchor-request-params-parser.js'
 import bodyParser from 'body-parser'
-import { isLeft } from 'fp-ts/lib/Either.js'
-import { makeErrorMessage } from '../ancillary/throw-decoder.js'
-import * as t from 'io-ts'
-import * as te from '../ancillary/io-ts-extra.js'
+import * as t from 'codeco'
```
dbcfd, line 16: Similar to previous comment, can we get rid of `*` import or name something else?

src/models/metadata.ts
```diff
@@ -1,21 +1,17 @@
-import * as t from 'io-ts'
-import * as te from '../ancillary/io-ts-extra.js'
+import * as t from 'codeco'
```
dbcfd, line 1: Same comment.

src/models/request.ts
```diff
@@ -1,5 +1,5 @@
-import * as t from 'io-ts'
-import * as te from '../ancillary/io-ts-extra.js'
+import * as t from 'codeco'
```
dbcfd, line 1: Same comment.

src/repositories/metadata-repository.ts
```diff
@@ -1,8 +1,8 @@
 import type { Knex } from 'knex'
 import { MetadataInput, StoredMetadata } from '../models/metadata.js'
-import { ThrowDecoder } from '../ancillary/throw-decoder.js'
 import type { StreamID } from '@ceramicnetwork/streamid'
-import * as te from '../ancillary/io-ts-extra.js'
+import * as t from 'codeco'
```
dbcfd, line 4: Same comment

src/repositories/metadata-repository.ts
```diff
@@ -47,6 +47,7 @@ export class MetadataRepository implements IMetadataRepository {
    * Store metadata entry to the database.
    */
   async save(entry: MetadataInput): Promise<void> {
+    console.log('save', entry, MetadataInput.encode(entry))
```
dbcfd, line 50: Should this be a logger call, or should it be removed?

dbcfd (changes requested):

src/ancillary/__tests__/anchor-request-params-parser.test.ts
ukstv: As `t` effectively is a namespace, the name should be short (it appears in `t.type({a: t.string})`) and have low probability of conflicting with other code. `t` was chosen as it is traditionally what `io-ts` recommends. Capital `C`? Or `CDC`? Suggestions are welcome.

src/ancillary/codecs.ts
ukstv: Oh, here it is! The suggestion! Accepted.

src/repositories/metadata-repository.ts
ukstv: Dang. Yes. Thank you for noticing this.

ukstv:
@dbcfd I just replaced all `import * as ...` statements with concrete imports. Please, take a look.

PR #1033, opened 2023-05-15
add metrics for car and legacy requests

gvelez17:
# add metrics for car and legacy requests - #PLAT-1496
## Description

This was a requested in a meeting, the metrics should not be able to break anything


ukstv (approved):

ukstv:
should be merged IMO

PR #1037, opened 2023-05-17
feat: replace ethers with viem

dbcfd:


src/models/transaction.ts
```diff
@@ -1,13 +1,11 @@
 export class Transaction {
   chain: string
   txHash: string
-  blockNumber: number
-  blockTimestamp: number
+  blockNumber: bigint
```
ukstv, line 4: Mind if I ask you what happened to the timestamp?

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
```diff
@@ -141,21 +161,17 @@ export class EthereumBlockchainService implements BlockchainService {
...
   private readonly overrideGasConfig: boolean
-  private readonly gasLimit: number
-  private readonly useSmartContractAnchors: boolean
-  private readonly contractAddress: string
+  private readonly gasLimit: bigint
+  private readonly contractAddress: Address
 
-  constructor(config: Config, private readonly wallet: ethers.Wallet) {
-    this.useSmartContractAnchors = config.useSmartContractAnchors
+  constructor(config: Config, private readonly provider: PublicClient, private readonly wallet: WalletClient, private readonly walletAddress: Address) {
```
ukstv, line 168: It is safer IMO to not mix `private readonly` constructor properties with property declarations like on 161-166. It makes it complicated to see what is actually available. IMO.

src/models/transaction.ts
dbcfd: It's not part of the transaction receipt from viem. Still need to figure out if there's a way for me to get that. Possibly by grabbing the block.

src/services/blockchain/ethereum/ethereum-blockchain-service.ts
dbcfd: I agree, will adjust that.

ukstv:
Hey @dbcfd any news on the PR?

PR #1038, opened 2023-05-17
feat: add origin ip and did when available to anchor err

zachferland:
follow up from discord

Doesn't really do too much, since if didnt have IP before it wasnt in the allow list anyways, but does make cloudwatch queries a bit easier/more consistent and may be useful at times


stbrody (approved):
lgtm

PR #1041, opened 2023-05-22
add metrics around eth transaction requests and errors

gvelez17:

## Description

## How Has This Been Tested?

- [ ] Test B

## Definition of Done

## References:

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -36,7 +36,7 @@ export enum METRIC_NAMES {
   RETRY_FAILED = 'retry_failed',
 
   // *******************************************************************//
-  // Anchor Service (histograms)
+  // Anchor Service (several of these are histograms)
 
   // request that moves from ready -> processing
   READY_PROCESSING_MS = "ready_processing_ms",
@@ -47,6 +47,13 @@ export enum METRIC_NAMES {
   // Anchor Controller
   ANCHOR_REQUESTS_BATCH_TIME = 'anchor_requests_batch_time',
   ANCHOR_REQUESTS_BATCH_FAILURE_TIME = 'anchor_requests_batch_failure_time',
+
+
+  // *******************************************************************//
+  // Ethereum Blockchain Service
+
+  ETH_REQUEST_TOTAL = 'eth_request_total',
+  ETH_REQUEST_ERROR_TOTAL = 'eth_request_error_total',
 
   // *******************************************************************//
   // Scheduler Service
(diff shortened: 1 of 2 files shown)


src/settings.ts
```diff
@@ -48,6 +48,13 @@ export enum METRIC_NAMES {
   ANCHOR_REQUESTS_BATCH_TIME = 'anchor_requests_batch_time',
   ANCHOR_REQUESTS_BATCH_FAILURE_TIME = 'anchor_requests_batch_failure_time',
 
+
+  // *******************************************************************//
+  // Ethereum Blockchain Service
+
+  ETH_REQUEST_TOTAL = 'eth_request_total',
```
stephhuynh18, line 55: What does request stand for? Based on the code it seems that it would be equal to the number of anchors we attempt. Is that what we want?

gvelez17: @stephhuynh18 so the request was to track the number of requests to the ethereum endpoint (and added on errors also).  That probably does match the number of anchor batches we attempt, tho not sure if it matches the total of individual anchors - don't we batch them into eth requests?  

The concern here was to track eth requests and failures, also velocity, i think this came up when we ran out of funds in the wallet

Since the concern is with the eth wallet maybe we shoudl track here, so if we change the batching logic, we still will track specifically the eth requests?

gvelez17: @stbrody this was originally your request i think, is that a good summary of why? ^

stephhuynh18: per anchor we hit the ethereum endpoint many times (get balance, get nonce, send raw transaction). Are we trying to track each individual call. Also are counting the number of requests where it is now, does not take into account the retries

gvelez17: i'm not sure what the exact purpose was actually.  @stbrody do you know what the exact desired metric is?  each time we hit the endpoint? do we want to count the retries?  I'll ask this in the ticket as well

stbrody: aww geez, this request was from over 3 months ago now so I don't really remember the context.  Looking at the ticket, the motivating factor seemed to be we were hitting infura rate limits.  I mean, I guess in an ideal world we'd have a separate metric to track each type of call we make to the eth rpc endpoint (chainid, get balance, get nonce, send transaction, etc), then we could sum up all those separate metrics to see the overall total number of requests of any type.

gvelez17: ok, i think maybe one metric with multiple labels (a label for each type of call) would be right then

PR #1044, opened 2023-05-24
fix: add payload limit to raw body parser

smrz2001:
Tested manually

### DIFF
--- a/auth/src/handlers/api.ts
+++ b/auth/src/handlers/api.ts
@@ -20,8 +20,7 @@ const email = new SESService()
 const gateway = new ApiGateway()
 const metrics = new CloudMetrics()
 
-// limit increased from default 100kb to accommodate CAR file payloads
-app.use(bodyParser.json({limit: '1mb'}))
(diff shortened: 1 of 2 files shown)


willex (approved):
🔥

auth/src/handlers/api.ts
```diff
@@ -20,8 +20,7 @@ const email = new SESService()
 const gateway = new ApiGateway()
 const metrics = new CloudMetrics()
 
-// limit increased from default 100kb to accommodate CAR file payloads
-app.use(bodyParser.json({limit: '1mb'}))
```
stbrody, line 24: so what's this parser used for?  Is this one used for regular commits but not for CAR file data?  Commits can be up to 256kb so the default 100kb won't be enough for those either.

nathanielc: Also, this app is the `auth` app, which only gets basic metadata about the request in order to authorize it. That is the part we missed on first review.

PR #1045, opened 2023-05-26
feat: Create Merkle CAR file while creating an anchor

ukstv:
And remove `includeBlockInfoInAnchorProof` and timestamp information from an anchor proof.


ukstv:
Wrong branch.

PR #1046, opened 2023-05-26
feat: Create Merkle CAR file while creating an anchor

ukstv:
And remove includeBlockInfoInAnchorProof and timestamp information from an anchor proof


src/merkle/merkle-car-factory.ts
```diff
@@ -0,0 +1,90 @@
...
+export interface IMerkleTree<TData, TLeaf extends TData, TMetadata> {
+  readonly root: Node<TData>
+  readonly leafNodes: Array<Node<TLeaf>>
+  readonly metadata: TMetadata | null
+  getProof(elemIndex: number): Array<Node<TData>>
+  getDirectPathFromRoot(elemIndex: number): PathDirection[]
+  verifyProof(proof: Node<TData>[], element: TData): Promise<boolean>
+}
+
+export class MerkleCAR implements IMerkleTree<CIDHolder, Candidate, TreeMetadata> {
```
stbrody, line 46: why do we need this class?  It doesn't really seem to do anything but associate a CAR with a MerkleTree - all the functionality though is a direct pass-through to the underlying merkle tree.  Would it be simpler to just pass around a struct with both the tree and the car file?

src/services/anchor-service.ts
```diff
@@ -393,27 +334,19 @@ export class AnchorService {
    * @param tx - ETH transaction
    * @param merkleRootCid - CID of the root of the merkle tree that was anchored in 'tx'
    */
-  async _createIPFSProof(tx: Transaction, merkleRootCid: CID): Promise<CID> {
-    const txHashCid = Utils.convertEthHashToCid(tx.txHash.slice(2))
-    let ipfsAnchorProof = {
+  _createIPFSProof(car: CAR, tx: Transaction, merkleRootCid: CID): CID {
```
stbrody, line 337: doc comment is out of date now and needs updating

src/services/ipfs-service.ts
```diff
@@ -155,4 +166,39 @@ export class IpfsService implements IIpfsService {
 
     return anchorCid
   }
+
+  async importCAR(car: CAR, options: AbortOptions = {}): Promise<void> {
+    const blocks = []
+    for (const block of car.blocks) {
+      blocks.push(block)
+    }
+    await Promise.all(blocks.map(async block => {
```
stbrody, line 175: does this risk overwhelming ipfs with too many simultaneous calls to `block.put`?  Maybe we should batch this or even just do them all one at a time since latency isn't that important here anyway?

src/services/ipfs-service.ts
```diff
@@ -155,4 +166,39 @@ export class IpfsService implements IIpfsService {
...
+      const cid = block.cid
+      const format = this.codecNames.get(cid.code)
+      const mhtype = this.hasherNames.get(block.cid.multihash.code)
+      await this.ipfs.block.put(block.payload, {
+        format: format,
+        mhtype: mhtype,
+      })
+      await this.ipfs.pin.add(block.cid, {
```
stbrody, line 183: ooh I didn't think to pin the blocks in the ceramic-side PR.  Good call, I'll add that to mine

src/services/ipfs-service.ts
```diff
@@ -155,4 +166,39 @@ export class IpfsService implements IIpfsService {
...
+        signal: options.signal,
+        timeout: IPFS_PUT_TIMEOUT,
+        recursive: false,
+      })
+    }))
+    // for (const block of car.blocks) {
```
stbrody, line 189: reminder to delete these commented lines before merge

src/merkle/merkle-car-factory.ts
stbrody: Nevermind, I see now.  It overrides the IPFSService to ensure that as the tree is being built each intermediate node in the tree is added to the CAR file.  That behavior is a bit subtle, some more comments throughout this file could help.

stbrody (commented):
Mostly looks good.  My main request is for a unit test that when doing an anchor the right data gets into the CAR file.

stbrody (approved):
lgtm

PR #1047, opened 2023-05-27
chore: Clearer Anchor code

ukstv:


src/models/anchor.ts
```diff
@@ -1,9 +1,19 @@
...
-  path!: string
-  cid!: string
-  proofCid!: string
-  createdAt!: Date
-  updatedAt!: Date
-}
+import { type, string, TypeOf } from 'codeco'
+import { cidAsString, date } from '../ancillary/codecs.js'
+
+export const FreshDatabaseAnchor = type({
```
stbrody, line 4: What do you think about calling this type `UnpersistedAnchor`?  I feel like it's a bit confusing calling it a `DatabaseAnchor` when its not yet in the database

stbrody (approved):

src/models/anchor.ts
ukstv: I changed the names similar to how they are for metadata. We have `StoredMetadata` and `FreshMetadata`. Similarly, now these two are called `StoredAnchor` and `FreshAnchor`. Is it all right for you?

stbrody: yeah that sounds good, thanks

PR #1062, opened 2023-06-06
feat: add failure queue

stephhuynh18:
Not using yet but we can post partial failures here at a later time

### DIFF
--- a/src/services/queue/sqs-queue-service.ts
+++ b/src/services/queue/sqs-queue-service.ts
@@ -4,20 +4,30 @@ import {
   Message,
   DeleteMessageCommand,
   ChangeMessageVisibilityCommand,
+  SendMessageCommand,
 } from '@aws-sdk/client-sqs'
-import { IQueueService, IQueueMessage } from './queue-service.type.js'
+import { QueueMessageData } from '../../models/queue-message.js'
+import {
+  IQueueConsumerService,
+  IQueueProducerService,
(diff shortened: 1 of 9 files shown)


src/services/queue/sqs-queue-service.ts
```diff
@@ -105,10 +126,37 @@ export class SqsQueueService<TInput, TValue> implements IQueueService<TValue> {
...
+  }
 }
 
-export class AnchorBatchSqsQueueService<TInput> extends SqsQueueService<TInput, AnchorBatch> {
+/**
+ * FailureSqsQueueService is used to consume and publish any failures that could happen during anchoring by the anchor workers.
+ * ex. A batch was partially completed so we can publish the partially complete batch to the failure queue.
+ * The failure queue can then handle it accordingly (such as putting the failed requests into a new batch) and provide alerts.
+ */
+export class FailureSqsQueueService extends SqsQueueService<QueueMessageData> {
```
smrz2001, line 158: We're not using this yet, right? Just the queue implementation?

smrz2001 (approved):
🥳

src/services/queue/sqs-queue-service.ts
stephhuynh18: yup exactly. I thought about it but nack is working sufficiently until we get into more complex failure handing situations

PR #1063, opened 2023-06-07
feat: add service for storing merkle tree CAR files to S3

stbrody:
Not yet tested


src/services/merkle-car-service.ts
```diff
@@ -0,0 +1,56 @@
...
+    return this.cars.get(anchorProofCID.toString()) || null
+  }
+}
+
+const S3_STORE_PATH = '/ceramic/cas/anchor/merkle-car/'
+const carFactory = new CARFactory()
+carFactory.codecs.add(DAG_JOSE)
+
+export class S3MerkleCarService {
+  readonly s3store: LevelUp.LevelUp = new LevelUp(new S3LevelDOWN(S3_STORE_PATH, new AWSSDK.S3()))
```
ukstv, line 35: How is it configured? I mean, I would expect some credentials and a name of s3 bucket. Does that happens automagically somehow?

stbrody: Yeah, my understanding is that the S3 library knows to look for certain specific environment variables and gets its configuration from that.  I believe that's how we usually configure the S3 settings for our Ceramic nodes currently.  @smrz2001 @3benbox can you confirm?

stbrody:
Added @smrz2001 to the review to check out the S3 configuration

config/env/test.json
```diff
@@ -5,6 +5,10 @@
   "readyRetryIntervalMS": 10000,
   "requireAuth": false,
   "schedulerIntervalMS": 10000,
+  "carStorage": {
+    "mode": "s3",
+    "s3BucketName": "myS3Bucket"
```
stbrody, line 10: is there a better default to use here?  Maybe we should put whatever we are actually using in our infra?  @smrz2001

src/services/merkle-car-service.ts
```diff
@@ -0,0 +1,64 @@
...
+  async retrieveCarFile(anchorProofCID: CID): Promise<CAR | null> {
+    return this.cars.get(anchorProofCID.toString()) || null
+  }
+}
+
+const S3_STORE_SUFFIX = '/cas/anchor/merkle-car/'
+const carFactory = new CARFactory()
+carFactory.codecs.add(DAG_JOSE)
+
+export class S3MerkleCarService {
```
ukstv, line 35: @stbrody Supposedly this implements `IMerkleCarService`

src/app.ts
```diff
@@ -38,6 +38,7 @@ import { AnchorRequestParamsParser } from './ancillary/anchor-request-params-par
 import { HealthcheckService, IHealthcheckService } from './services/healthcheck-service.js'
 import { RequestService } from './services/request-service.js'
 import { AnchorBatchSqsQueueService } from './services/queue/sqs-queue-service.js'
+import { InMemoryMerkleCarService, S3MerkleCarService } from './services/merkle-car-service.js'
 
 type DependenciesContext = {
   config: Config
```
ukstv, line 44: Please, add `IMerkleCarService` to `DependenciesContext` @stbrody

config/env/test.json
smrz2001: Yeah, we use `ceramic-${env}-cas` so we could use `ceramic-dev-cas` here.

config/default.json
```diff
@@ -14,6 +14,10 @@
   "readyRetryIntervalMS": 300000,
   "requireAuth": false,
   "schedulerIntervalMS": 300000,
+  "carStorage": {
+    "mode": "inmemory",
+    "s3BucketName": "myS3Bucket"
```
smrz2001, line 19: Can also use the default name suggested below.

Is it possible to parameterize this and use something like `ceramic-${env}-cas` here?

src/app.ts
ukstv: okay, disregard this one. Does not matter really.

src/services/merkle-car-service.ts
ukstv: Again, please disregard the comment.

PR #1071, opened 2023-06-12
chore: use DATABASE_URL for the db connection if set in tests

stephhuynh18:
Currently if you would like to use a local db for tests, it has to have a specific configuration (test-db, test-user, etc. Now you can use the db url you provided with any configuration you want.

### DIFF
--- a/src/db-connection.ts
+++ b/src/db-connection.ts
@@ -41,7 +41,13 @@ async function runMigrations(connection: Knex) {
 
 export async function createDbConnection(dbConfig: Db = config.db): Promise<Knex> {
   const knexConfig: Knex.Config = {
-    ...dbConfig,
+    client: dbConfig.client,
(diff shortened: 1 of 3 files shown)


src/db-connection.ts
```diff
@@ -41,7 +41,13 @@ async function runMigrations(connection: Knex) {
 
 export async function createDbConnection(dbConfig: Db = config.db): Promise<Knex> {
   const knexConfig: Knex.Config = {
-    ...dbConfig,
+    client: dbConfig.client,
+    connection: dbConfig.connection.connectionString
```
ukstv, line 45:
```suggestion
    connection: dbConfig.connection.connectionString || dbConfig.connection
```

If I am not mistaken, this achieves the same result, yet it is less complicated to read.

ukstv (approved):

PR #1072, opened 2023-06-12
feat: Add graceful shutdown to scheduling service

stephhuynh18:
This was not specifically for a CAS in "scheduler" mode but useful


smrz2001 (approved):
🥳

PR #1075, opened 2023-06-14
feat: continual anchoring mode

stephhuynh18:
tested via integration tests in go-cas


src/app.ts
```diff
@@ -78,7 +79,8 @@ export class CeramicAnchorApp {
     this.anchorsSupported =
       this.mode === AppMode.ANCHOR ||
       this.mode === AppMode.BUNDLED ||
-      this.config.anchorControllerEnabled
+      this.mode === AppMode.CONTINUAL_ANCHORING
```
smrz2001, line 82: Is this a typo? Looks like the `||` is missing at the end of this line.

stephhuynh18: yup!

smrz2001 (approved):
Not super fluent with the observable stuff but mostly LGTM! 🚢

PR #1076, opened 2023-06-15
feat: Send CAR file, and use codecs

ukstv:
Treat that as a work in progress for now.

Ah, it supposedly fixes AWS S3 initialisation bug.


src/services/request-presentation-service.ts
```diff
@@ -1,99 +1,135 @@
...
-   */
-  async body(request: Request): Promise<RequestPresentation> {
+  constructor(
+    private readonly anchorRepository: IAnchorRepository,
+    private readonly merkleCarService: IMerkleCarService,
+    private readonly witnessService: WitnessService
+  ) {}
+
+  async witnessCAR(anchor: AnchorWithRequest | null): Promise<CAR | null> {
+    if (!anchor) return null // TODO Add metric
```
stbrody, line 46: can we file a ticket for this todo and reference the ticket number here?

src/services/request-presentation-service.ts
```diff
@@ -1,99 +1,135 @@
...
+    const cacheKey = anchor.cid.toString()
+    const fromCache = this.cache.get(cacheKey)
+    if (fromCache) return fromCache
+    const merkleCAR = await this.merkleCarService.retrieveCarFile(anchor.proofCid)
+    if (!merkleCAR) return null // TODO Add metric
```
stbrody, line 51: we should probably log a warning for this case as it isn't expected

src/services/request-presentation-service.ts
```diff
@@ -1,99 +1,135 @@
...
     switch (status) {
       case RequestStatus.COMPLETED: {
         const anchor = await this.anchorRepository.findByRequest(request)
+        const witnessCAR = await this.witnessCAR(anchor)
         // TODO: This is a workaround, fix in CDB-2192
-        const anchorCommit = {
-          cid: anchor ? anchor.cid.toString() : request.cid,
+        const anchorCommit: CommitPresentation = {
+          cid: anchor ? anchor.cid : CID.parse(request.cid),
           content: {
```
stbrody, line 66: not for this PR, but as a general question - should we remove this content object from the response entirely if it's not being used anywhere?

stbrody (approved):
LGTM!

src/services/request-presentation-service.ts
ukstv: Better do that here :D

src/services/request-presentation-service.ts
ukstv:
```ts
logger.warn(`No Merkle CAR found for anchor ${anchor.cid}`)
```

src/services/request-presentation-service.ts
ukstv: Yep. That was done in js-ceramic codec. And, removed `content` from CAS codebase as well.

PR #1083, opened 2023-06-19
fix: Ignore replaced requests when anchoring batches

stephhuynh18:
We do not want to anchor replaced requests. This will reduce the batch size. But will ensure the latest commit will be anchored.


smrz2001 (approved):
:shipit: 🚀

PR #1084, opened 2023-06-19
feat: Creating requests publishes to validation queue

stephhuynh18:
Instead of marking requests replaced in the controller, we send requests to the validation service which will provide the deduping algorithm. This will reduce the concurrent updates.


src/services/request-service.ts
```diff
@@ -10,20 +10,30 @@ import type { RequestAnchorParams } from '../ancillary/anchor-request-params-par
...
   constructor(
+    config: Config,
     private readonly requestRepository: RequestRepository,
     private readonly requestPresentationService: RequestPresentationService,
-    private readonly metadataService: IMetadataService
-  ) {}
+    private readonly metadataService: IMetadataService,
+    private readonly validationQueueService: IQueueProducerService<RequestQMessage>
+  ) {
+    this.publishToQueue = config.useQueueBatches && config.queue.sqsQueueUrl !== ''
```
smrz2001, line 35: Didn't we want to use just the presence or absence of the queue URL to decide v5 mode? I took `useQueueBatches` out of the config but I can add it back if preferable.

stephhuynh18: oops missed that! will fix

smrz2001 (approved):
LGTM apart from the minor change to the config! 🥳

PR #1088, opened 2023-06-23
temporarily disable witness CAR construction

stbrody:


stbrody:
we fixed the issue with S3 access perms, this is not necessary

PR #1096, opened 2023-07-02
test: Integration test CAR-Ceramic

ukstv:
Check if a Ceramic node gets an anchor from witness CAR passed from CAS.

Also, had to pin OpenZeppelin version so that the tests pass. They updated to a solc version to newer than ours.

### DIFF
--- a/src/services/anchor-service.ts
+++ b/src/services/anchor-service.ts
@@ -476,6 +476,9 @@ export class AnchorService {
 
     try {
       const anchorCid = await this.ipfsService.storeRecord(ipfsAnchorCommit)
+      logger.debug(
+        `Stored anchor commit for ${candidate.cid} of stream ${candidate.streamId} on IPFS`
+      )
       await this.ipfsService.publishAnchorCommit(anchorCid, candidate.streamId)
       const anchor: FreshAnchor = {
         requestId: candidate.request.id,
--- a/src/services/anchor-service.ts
+++ b/src/services/anchor-service.ts
@@ -472,13 +472,11 @@ export class AnchorService {
       proof: ipfsProofCid,
       path: path,
     }
+    car.put(ipfsAnchorCommit)
 
     try {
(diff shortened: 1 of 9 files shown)


src/services/ipfs-service.ts
```diff
@@ -142,29 +142,25 @@ export class IpfsService implements IIpfsService {
 
   /**
    * Stores `anchorCommit` to ipfs and publishes an update pubsub message to the Ceramic pubsub topic
```
stephhuynh18, line 144: nitpick: change the comment too

stephhuynh18, line 144: nitpick: change the comment too

stephhuynh18 (approved):

src/services/anchor-service.ts
```diff
@@ -495,10 +509,7 @@ export class AnchorService {
       } for stream ${candidate.streamId.toString()}: ${err}`
       logger.err(msg)
       Metrics.count(METRIC_NAMES.ERROR_IPFS, 1)
-      await this.requestRepository.updateRequests({ status: RS.FAILED, message: msg }, [
```
stbrody, line 498: this change is interesting.  I think it's probably fine and makes sense though - even if the anchor doesn't get to IPFS it's data can successfully get to Ceramic via the CAR file so that's fine.

So is the behavior now that the anchor is marked as COMPLETED even if the publish to ipfs/pubsub fails?

stbrody (commented):
lgtm!  Exciting.

src/services/anchor-service.ts
ukstv: Yes. As you know, passing witness CAR is part of a bigger attempt to eliminate excessive reliance on IPFS/pubsub. This change makes it explicit. If you can not announce an anchor over pubsub, or can not store a record on IPFS, it should not be a showstopper for the anchoring process.

PR #1104, opened 2023-07-05
feat: Failed status is terminal

ukstv:
Nack a whole batch if failed to store a Merkle CAR to S3 or IPFS, log an error.

### DIFF
--- a/src/services/__tests__/anchor-service.test.ts
+++ b/src/services/__tests__/anchor-service.test.ts
@@ -17,7 +17,7 @@ import { CID } from 'multiformats/cid'
 import type { FreshAnchor } from '../../models/anchor.js'
 import { toCID } from '@ceramicnetwork/common'
 import { Utils } from '../../utils.js'
-import { validate as validateUUID, v4 as uuidv4 } from 'uuid'
+import { v4 as uuidv4, validate as validateUUID } from 'uuid'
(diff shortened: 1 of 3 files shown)


src/services/__tests__/anchor-service.test.ts
```diff
@@ -802,5 +728,53 @@ describe('anchor service', () => {
...
+        return Promise.resolve(fakeTransaction)
+      }
+
+      jest.spyOn(ipfsService, 'importCAR').mockImplementation(async () => {
+        throw new Error(`Can not import merkle CAR`)
+      })
+      await expect(anchorService.anchorRequests()).rejects.toThrow()
+      const retrieved = await requestRepository.findByIds(requests.map((r) => r.id))
+      expect(retrieved.every((r) => r.status === RequestStatus.PENDING)).toBeTruthy()
+      blockchainService.sendTransaction = original
```
stephhuynh18, line 753: is a try finally block not needed?

stephhuynh18 (approved):
Just a small question

src/services/__tests__/anchor-service.test.ts
ukstv: There is no error thrown here, so I don't think "finally" is needed.

PR #1107, opened 2023-07-06
fix: do not assume default sqs queue url

stephhuynh18:
It is very weird that in kubernetes the default config does not populate sqsQueueUrl

### DIFF
--- a/src/services/anchor-service.ts
+++ b/src/services/anchor-service.ts
@@ -160,7 +160,7 @@ export class AnchorService {
   ) {
     this.merkleDepthLimit = config.merkleDepthLimit
     this.useSmartContractAnchors = config.useSmartContractAnchors
-    this.useQueueBatches = config.queue.sqsQueueUrl !== ''
+    this.useQueueBatches = !config.queue.sqsQueueUrl
 
     const minStreamCount = Number(config.minStreamCount)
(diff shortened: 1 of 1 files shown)


smrz2001 (approved):
🥳

src/services/anchor-service.ts
```diff
@@ -160,7 +160,7 @@ export class AnchorService {
   ) {
     this.merkleDepthLimit = config.merkleDepthLimit
     this.useSmartContractAnchors = config.useSmartContractAnchors
-    this.useQueueBatches = config.queue.sqsQueueUrl !== ''
+    this.useQueueBatches = Boolean(config.queue.sqsQueueUrl)
```
3benbox, line 163: I think this is due to javascript's truthyness vs Go?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

Either way, LGTM!

3benbox (approved):
LGTM

src/services/anchor-service.ts
stephhuynh18: I'm actually assuming that if the sqsQueueUrl isn't set, the default config will fill it in with an empty string. Thats why the issues you and Mohsin are seeing are so weird....

PR #1112, opened 2023-07-10
chore: Update README to include build step

stbrody:


stbrody:
Going ahead and merging without review due to triviality

PR #1114, opened 2023-07-13
please log all write requests for now, may remove this logging later

gvelez17:
# Verify logs of writes in CAS

## Description

Adding logging for createOrUpdate to verify that the requests are being created.

## How Has This Been Tested?

- [ ] Unit tests - this is a tiny change

### DIFF
--- a/src/services/request-service.ts
+++ b/src/services/request-service.ts
@@ -99,7 +99,7 @@ export class RequestService {
       family: genesisFields?.family,
       model: genesisFields?.model,
     })
-    logger.debug(`Write requested: ${request.cid}`)
+    logger.imp(`Write requested: ${request.cid}`)
 
     return this.requestPresentationService.body(storedRequest)
   }


src/settings.ts
```diff
@@ -79,3 +79,4 @@ export enum METRIC_NAMES {
   REQUEST_INVALID = 'request_invalid',
   REQUEST_NOT_CREATED = 'request_not_created',
 }
+
```
gvelez17, line 82: this was just to trigger the tests to rerun when it had a weird codecov error and wouldn't rerun

src/services/request-service.ts
```diff
@@ -99,7 +99,7 @@ export class RequestService {
       family: genesisFields?.family,
       model: genesisFields?.model,
     })
-    logger.debug(`Write requested: ${request.cid}`)
+    logger.imp(`Write requested: ${request.cid}`)
```
dbcfd, line 102: Is the log level too high if we just set the logger to log at debug level?

dbcfd (approved):

src/services/request-service.ts
stbrody, line 102: can we make clearer what "write" we're talking about here?

Maybe something like `"Anchor request received for CID ${request.cid} of streamid ${request.streamId} from origin ${request.origin}"`?

stbrody (approved):
lgtm, with a suggested re-wording of the message to be a bit more detailed

src/services/request-service.ts
gvelez17: yes - my intention is to remove this log later but yes good idea

PR #1115, opened 2023-07-14
fix: Store bytes to S3

ukstv:
I fucked up. We try to store `Uint8Array` to S3 which [gets turned into String](https://github.com/loune/s3leveldown/blob/master/s3leveldown.js#L240) of numbers like `58,62,101,114, ...`, like what happens when you do `String(uint8array)`. The PR amends that.


stbrody (approved):
interesting.  If this is a problem why didn't we notice it before?  We do see some logs that suggest some CAR files are being successfully loaded from S3 - why does it work sometimes and not others?

I guess the integration test didn't catch this because it doesn't use S3 😕

ukstv:
It was there. The traffic was negligible on tnet.
Screenshot 2023-07-14 at 17 12 28
Screenshot 2023-07-14 at 17 12 57

PR #1125, opened 2023-07-26
feat: add max attempts when polling

stephhuynh18:
Default is 3

Also i noticed that for env vars, the default config does not get used therefore I've set default values in the queue service.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -76,7 +76,7 @@
     "sqsQueueUrl": "",
     "pollingIntervalMS": 10000,
     "maxAttemptsToRetrieveMessages": 3,
-    "maxTimeToHoldMessageSec": 10800,
-    "waitTimeForMessageSec": 30
+    "maxTimeToHoldMessageSec": 21600,
+    "waitTimeForMessageSec": 0
   }
 }
--- a/config/default.json
+++ b/config/default.json
@@ -75,6 +75,7 @@
     "awsRegion": "us-east-1",
     "sqsQueueUrl": "",
     "pollingIntervalMS": 10000,
+    "maxAttemptsToRetrieveMessages": 3,
     "maxTimeToHoldMessageSec": 10800,
     "waitTimeForMessageSec": 30
   }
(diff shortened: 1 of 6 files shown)


src/services/queue/sqs-queue-service.ts
```diff
@@ -18,6 +18,10 @@ import { Codec, decode } from 'codeco'
 import { AbortOptions } from '@ceramicnetwork/common'
 import { Utils } from '../../utils.js'
 
+const DEFAULT_POLLING_INTERVAL_MS = 10000
+const DEFAULT_MAX_TIME_TO_HOLD_MESSAGES_S = 10800
```
smrz2001, line 22: Should this be 6 hours since that's what we use for Prod in go-cas?

config/default.json
```diff
@@ -75,6 +75,7 @@
     "awsRegion": "us-east-1",
     "sqsQueueUrl": "",
     "pollingIntervalMS": 10000,
+    "maxAttemptsToRetrieveMessages": 3,
     "maxTimeToHoldMessageSec": 10800,
     "waitTimeForMessageSec": 30
```
smrz2001, line 80: Should we set this to 0?

The wait time is usually meant to allow messages to collect so that SQS batches can be fuller and fewer API calls are needed overall.

A worker only picks up one job at a time (so it doesn't care about waiting for fuller SQS batches), and it's only started when there is >= 1 job to process (so we know it'll almost always have work to do).

config/env/test.json
```diff
@@ -58,6 +58,7 @@
     "awsRegion": "us-east-1",
     "sqsQueueUrl": "",
     "pollingIntervalMS": 10000,
+    "maxAttemptsToRetrieveMessages": 0,
```
smrz2001, line 61: This will cause workers to wait forever? That's ok to do for tests?

stephhuynh18: The tests should timeout before then. Prior to this, the attempt were unlimited.

PR #1126, opened 2023-07-28
feat: Additional logging when accepting and serving requests

ukstv:


stbrody (approved):
I'm a little worried this will be too verbose and fill up our logs and make them hard to read, but we can always remove logs or up their log level if we need to after we've figured out what we're hoping to learn, so fine with putting this in at least for now

ukstv:
@stbrody It's temporary to see how often CAS gets requests for old commits.

PR #1127, opened 2023-07-28
feat: stop continual anchoring if we receive an empty batch

stephhuynh18:
Please use environment vars:

APP_MODE=continual-anchoring
SCHEDULER_STOP_AFTER_FAIL=true

to turn this on

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -132,7 +132,7 @@ export class CeramicAnchorApp {
 
   async anchor(): Promise<void> {
     const anchorService = this.container.resolve('anchorService')
-    return anchorService.anchorRequests()
+    await anchorService.anchorRequests()
   }
 
   /**
@@ -286,12 +286,13 @@ export class CeramicAnchorApp {
     process.on('SIGTERM', shutdownSignalHandler)
     process.on('SIGQUIT', shutdownSignalHandler)
 
-    const task = async () => {
-      await anchorService.anchorRequests({ signal: controller.signal }).catch((error) => {
+    const task = async (): Promise<boolean> => {
+      const success = await anchorService.anchorRequests({ signal: controller.signal }).catch((error: Error) => {
         logger.err(`Error when anchoring: ${error}`)
         Metrics.count(METRIC_NAMES.ERROR_WHEN_ANCHORING, 1, {
           message: error.message.substring(0, 50),
         })
+        throw error
       })
 
       logger.imp(
@@ -301,8 +302,16 @@ export class CeramicAnchorApp {
       // await anchorService.garbageCollectPinnedStreams().catch((error) => {
       //   logger.err(`Error when garbage collecting pinned streams: ${error}`)
       // })
+
+      return success
     }
 
-    continualAnchoringScheduler.start(task, this.config.schedulerIntervalMS)
+
+    const cbAfterEmptyBatch = () => {
+      logger.imp('No batches available. Continual anchoring is shutting down.')
+      process.exit(0)
+    }
+
+    continualAnchoringScheduler.start(task, this.config.schedulerIntervalMS, this.config.schedulerStopAfterFail ? cbAfterEmptyBatch : undefined)
   }
 }
--- a/src/app.ts
+++ b/src/app.ts
@@ -307,11 +307,11 @@ export class CeramicAnchorApp {
     }
 
 
-    const cbAfterEmptyBatch = () => {
+    const cbAfterNoOp = () => {
       logger.imp('No batches available. Continual anchoring is shutting down.')
       process.exit(0)
     }
 
-    continualAnchoringScheduler.start(task, this.config.schedulerIntervalMS, this.config.schedulerStopAfterFail ? cbAfterEmptyBatch : undefined)
+    continualAnchoringScheduler.start(task, this.config.schedulerIntervalMS, this.config.schedulerStopAfterNoOp ? cbAfterNoOp : undefined)
   }
 }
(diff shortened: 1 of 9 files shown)


src/app.ts
```diff
@@ -301,8 +302,16 @@ export class CeramicAnchorApp {
...
       //   logger.err(`Error when garbage collecting pinned streams: ${error}`)
       // })
+
+      return success
+    }
+
+
+    const cbAfterEmptyBatch = () => {
+      logger.imp('No batches available. Continual anchoring is shutting down.')
+      process.exit(0)
```
smrz2001, line 312: Do we usually call exit directly in the process like this?

src/services/task-scheduler-service.ts
```diff
@@ -47,22 +54,37 @@ export class TaskSchedulerService {
...
         return timer(intervalMS).pipe(concatMap(() => taskWithRetryOnError$))
       }),
       share()
     )
 
-    this._scheduledTask$.subscribe()
+
+    this._scheduledTask$.subscribe({
+      complete: async () => {
+        if (!this._completed) {
```
smrz2001, line 75: Does it matter that `_completed` could get set to `true` even if no `cbAfterFailure` was specified?

src/app.ts
stephhuynh18: Yes I'm following the conventions in this file (ex. https://github.com/ceramicnetwork/ceramic-anchor-service/blob/6a73c563c8fdecfe04f2064cd632c107e3dd9fd1/src/app.ts#L262C5-L262C19)

src/services/task-scheduler-service.ts
```diff
@@ -47,22 +54,37 @@ export class TaskSchedulerService {
...
     this._scheduledTask$ = taskWithRetryOnError$.pipe(
-      expand(() => {
+      expand((success: boolean) => {
         if (this._controller.signal.aborted) {
           return EMPTY
         }
 
+        if (cbAfterFailure && !success) {
+          logger.imp(`Last run of the task was not successful. Stopping the task scheduler`)
+          return EMPTY
```
smrz2001, line 64: Will this cause an exit after the very first time a batch isn't found? Or will it run through the extra attempts and then exit if each attempt fails?

smrz2001: It actually doesn't matter a lot if it exits after the first attempt. As long as we're making an attempt to check the queue before shutting down.

src/services/task-scheduler-service.ts
stephhuynh18: This was added because I found a bug when calling stop on the task scheduler. But this reminded me of the rxjs property `Subscription.closed` which makes this not needed.

src/services/task-scheduler-service.ts
stephhuynh18: So there are two "loops" in this situation.

1. The task is anchoring a queued batch. Anchoring a queued batch has a loop when retrieving a task. 
2. The task scheduler service which runs the task after a delay of 1000s 

Originally I added #1 because for our docker environment we wanted to run the workers in "anchor" mode and have the worker wait for a message. So I made waiting for a message run in a loop. 

I then added #2 because we wanted to do greedy anchoring. So we run #1 in a loop with a dealy

This didn't really work with our infra because #1 caused use to wait indefinitely for a message which meant long living workers. So i think i'm just going to remove #1 because its confusing, and rely on the task scheduler

smrz2001: Ok, I see, ty for explaining.

So, IIUC, now we're relying on the existing task/loop code to attempt to retrieve queue messages. With a 10 second wait time, the call to SQS will wait up to that long for a batch. If it finds one, it will process the batch, then look for more batches. If not, that will be considered a "failure" of which a max of 3 attempts will be permitted before a final failure is detected and the callback is invoked, exiting the process.

smrz2001: I definitely like the simplicity of the new approach!!

stephhuynh18: Almost. 

We are retrying on the existing task/loop code. This runs the task. The task is wait for 10s for a batch then:
1. if there is a batch anchor it. The task will then complete returning true. We then wait for a delay of schedulerIntervalMS before starting again
2. if there isn't a batch complete and return false indicating a no op. At that point the scheduler will invoke the cb exiting the process.

Its hard to get the the right terminology. The retry logic is only invoked when errors are thrown, not if the task returns false. If you have a better name please let me know

stephhuynh18: > And if we want to do long-lived workers in the future, we could make the retry count configurable and set it to infinity?

If we wanted long lived workers we would set schedulerStopAfterFail to false

smrz2001: Yeah, I like the differentiation between no-op and failure, otherwise it gets confusing.

smrz2001: > if there isn't a batch complete and return false indicating a no op. At that point the scheduler will invoke the cb exiting the process.

How does this retry factor in there?
```
      retry({
        delay: intervalMS,
        count: 3,
```

stephhuynh18: > > if there isn't a batch complete and return false indicating a no op. At that point the scheduler will invoke the cb exiting the process.
> 
> How does this retry factor in there?
> 
> ```
>       retry({
>         delay: intervalMS,
>         count: 3,
> ```

The retry is only if the task throws errors. Ex. The scheduler starts, the task throws an error, we retry. If it throws an error again we retry it etc. If we have retried it 3 times, the scheduler will throw causing the entire thing to stop

smrz2001 (approved):
Apart from renaming, LGTM! Thanks for taking care of this 🥳

PR #1152, opened 2023-08-30
fix: REPLACE all previous

ukstv:
Imagine a Ceramic node updates a stream with commit A and tries to request an anchor for commit A from the CAS, but the CAS is temporarily unavailable, say due to the network being down transiently.  The Ceramic node will then start a background task to retry creating the request once every 60 seconds until it goes though.  Then, 30 seconds later, the stream is updated again on the same node, with commit B.  By this point the network has recovered, and the request for commit B is successfully created against the CAS.  Then 30 seconds after that, the anchor request for commit A is retried and goes through.

We have a timestamp inside a request, so commit B still trumps commit A, but we have a bug (or maybe not @stephhuynh18 thanks to CASv5) as we consider REPLACED to be applied to previous requests _relative_ to the current one.

### DIFF
--- a/src/repositories/__tests__/request-repository.test.ts
+++ b/src/repositories/__tests__/request-repository.test.ts
@@ -878,9 +878,8 @@ describe('request repository test', () => {
         return requestRepository.createOrUpdate(request)
       })
       const requests: Array<Request> = await Promise.all(requestsP)
-      const first = requests[0]
-      expectPresent(first)
-      const rowsAffected = await requestRepository.markPreviousReplaced(first)
+      expectPresent(requests[0])
+      const rowsAffected = await requestRepository.markPreviousReplaced(requests[0])
       expect(rowsAffected).toEqual(requests.length - 1) // Mark every request except the last one
     })
   })
--- a/src/repositories/__tests__/request-repository.test.ts
+++ b/src/repositories/__tests__/request-repository.test.ts
@@ -795,8 +795,9 @@ describe('request repository test', () => {
   })
 
   describe('markPreviousReplaced', () => {
+    const ONE_HOUR_AGO = DateTime.fromISO('2023-01-01T00:00Z')
+
     test('mark older PENDING entries REPLACED', async () => {
-      const oneHourAgo = DateTime.fromISO('1900-01-01T00:00Z')
       const streamId = randomStreamID()
 
       // Create three COMPLETED requests. These should not be changed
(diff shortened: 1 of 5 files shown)


stephhuynh18 (approved):

stephhuynh18:
go-cas marks requests as replaced one at a time. It keeps track of the "most recent request" for a stream. If a request comes in and its older, we mark it as REPLACED. If the request is newer we mark the "most recent request" as REPLACED as the request becomes "the most recent request".  We never mark anything but PENDING requests as REPLACED

ukstv:
> go-cas marks requests as replaced one at a time. It keeps track of the "most recent request" for a stream. If a request comes in and its older, we mark it as REPLACED. If the request is newer we mark the "most recent request" as REPLACED as the request becomes "the most recent request". We never mark anything but PENDING requests as REPLACED

Aha, that's why @stbrody saw a proper REPLACED status. So, this PR basically replicates what you already do.

src/repositories/__tests__/request-repository.test.ts
```diff
@@ -863,6 +864,24 @@ describe('request repository test', () => {
...
+          timestamp: ONE_HOUR_AGO.plus({ minute: n }).toJSDate(),
+          status: RequestStatus.PENDING,
+          origin: 'same-origin',
+        })
+        return requestRepository.createOrUpdate(request)
+      })
+      const requests: Array<Request> = await Promise.all(requestsP)
+      expectPresent(requests[0])
+      const rowsAffected = await requestRepository.markReplaced(requests[0])
+      expect(rowsAffected).toEqual(requests.length - 1) // Mark every request except the last one
```
stbrody, line 883: can we check which requests get marked replaced not just the number of requests?  Otherwise it seems like if we did the date comparisons wrong and marked the wrong request as replaced this test would still pass

src/repositories/request-repository.ts
```diff
@@ -451,13 +453,21 @@ export class RequestRepository {
   }
 
   /**
-   * Mark PENDING requests older than `request.timestamp` REPLACED if they share same `request.origin` and `request.streamId`s.
+   * Mark all PENDING requests but the last one REPLACED if they share same `request.origin` and `request.streamId`s.
```
stbrody, line 456:
```suggestion
   * Mark all PENDING requests but the most recent one (by client-side creation timestamp, which may be different than the order that the CAS received them in) REPLACED if they share same `request.origin` and `request.streamId`s.
```

stbrody (approved):
lgtm

src/repositories/__tests__/request-repository.test.ts
ukstv: yup

PR #1157, opened 2023-09-05
feat: pubsub responder

stephhuynh18:
In the mode pubsub responder, it will listen to pubsub query messages and create an ipfs task to respond with the anchor commit

### DIFF
--- a/src/services/ipfs-service.ts
+++ b/src/services/ipfs-service.ts
@@ -203,74 +203,76 @@ export class IpfsService implements IIpfsService {
   }
 
   private async createPubsub() {
-    if (this.respondToPubsubQueries) {
-      const ipfsId = await this.ipfs.id()
-      this.pubsub$ = new Observable<Message>((subscriber) => {
-        const onMessage = (message: Message) => subscriber.next(message)
-        const onError = (error: Error) => subscriber.error(error)
-        this.ipfs.pubsub
-          .subscribe(this.pubsubTopic, onMessage, { onError })
-          .then(() => {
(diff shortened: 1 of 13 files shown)


src/services/ipfs-service.ts
```diff
@@ -174,4 +193,84 @@ export class IpfsService implements IIpfsService {
...
+  async stop(): Promise<void> {
+    if (this.pubsub$) {
+      this.pubsub$.unsubscribe()
+    }
+
+    await this.ipfs.pubsub.unsubscribe(this.pubsubTopic)
+  }
+
+  private async createPubsub() {
+    if (this.respondToPubsubQueries) {
```
stbrody, line 206: if you invert the condition here you can early return and avoid an indentation layer which will help with readability.

src/services/ipfs-service.ts
```diff
@@ -174,4 +193,84 @@ export class IpfsService implements IIpfsService {
...
+      // We have to subscribe to pubsub to keep ipfs connections alive.
+      // TODO Remove this when the underlying ipfs issue is fixed
+      await this.ipfs.pubsub.subscribe(this.pubsubTopic, () => {
+        /* do nothing */
+      })
+    }
+  }
+
+  async handleMessage(message: PubsubMessage): Promise<void> {
+    if (message.typ === MsgType.QUERY) {
```
stbrody, line 250: again, early return can avoid an indentation layer and increase readability.

stbrody (approved):
lgtm, but I'd like Sergey to take a look as well, especially at the rxjs plumbing

ukstv (approved):
Huge

PR #1162, opened 2023-09-11
Revert "move invalid carfile tests to separate PR"

gvelez17:
This reverts commit d82ee7bc22c5ae9f984b13b2d0140b04a8ca268b.

add test for invalid carfile having 'left' signal

### DIFF
--- a/src/ancillary/__tests__/anchor-request-params-parser.test.ts
+++ b/src/ancillary/__tests__/anchor-request-params-parser.test.ts
@@ -78,14 +78,6 @@ const CAR_FILE_INVALID = mockRequest({
 })
 
 
-const CAR_FILE_NOT_BASE64URL = mockRequest({
-  headers: {
-    'Content-Type': 'application/vnd.ipld.car',
-  },
-  body: '----------------------------------------',
-})
-
-
 const CAR_FILE_FAKE_GENESIS_FIELDS: GenesisFields = {
   controllers: [asDIDString('did:pkh:eip155:1:0x926eeb192c18b7be607a7e10c8e7a7e8d9f70742')],
   model: StreamID.fromBytes(
@@ -141,9 +133,4 @@ describe('AnchoRequestParamsParser', () => {
     expect(isLeft(validation)).toBeTruthy()
   })
 
-  test('throws error if cannot decode car file', () => {
-    const validation =  parser.parse(CAR_FILE_NOT_BASE64URL as ExpReq)
-    console.log("a string of dashes, turns to " + JSON.stringify(validation))
-    expect(() => parser.parse(CAR_FILE_NOT_BASE64URL as ExpReq)).toThrow(/^Can not decode CAR file/);
-  })
 })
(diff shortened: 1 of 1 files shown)


src/ancillary/__tests__/anchor-request-params-parser.test.ts
```diff
@@ -117,4 +128,9 @@ describe('AnchoRequestParamsParser', () => {
     expect(params.capCID).toEqual(FAKE_CAPCID)
   })
 
+  test('throws error on invalid car file', () => {
+    const validation =  parser.parse(CAR_FILE_INVALID as ExpReq)
+    console.log("Here is invalid validation: " + JSON.stringify(validation))
+    expect(() => parser.parse(CAR_FILE_INVALID as ExpReq)).toThrow(/^Can not decode CAR file/);
```
gvelez17, line 134: so, this doesn't throw an error, on the junk carfile.  instead it happily extracts this stuffs

```
Here is invalid validation: {"_tag":"Left","left":[{"trail":[{"key":"","type":{"name":"RequestAnchorParamsV2"},"actual":{"0":67,"1":114,"2":101,"3":97,"4":116,"5":101,"6":100,"7":66,"8":121,"9":67,"10":104,"11":97,"12":116,"13":71,"14":80,"15":84,"16":52,"17":89,"18":111,"19":117,"20":99,"21":97,"22":110,"23":85,"24":115,"25":101,"26":84,"27":104,"28":105,"29":115,"30":83,"31":116,"32":114,"33":105,"34":110,"35":103}}]}]}
```

gvelez17: @ukstv i moved it here so i could merge the other one, since this is on some old existing code

src/ancillary/__tests__/anchor-request-params-parser.test.ts
```diff
@@ -117,4 +136,14 @@ describe('AnchoRequestParamsParser', () => {
...
   })
 
+  test('isleft indicates invalid car file', () => {
+    const validation =  parser.parse(CAR_FILE_INVALID as ExpReq)
+    expect(isLeft(validation)).toBeTruthy()
+  })
+
+  test('throws error if cannot decode car file', () => {
+    const validation =  parser.parse(CAR_FILE_NOT_BASE64URL as ExpReq)
+    console.log("a string of dashes, turns to " + JSON.stringify(validation))
```
gvelez17, line 146: so validation of '---------' results in 

```
{"_tag":"Left","left":[{"trail":[{"key":"","type":{"name":"RequestAnchorParamsV2"},"actual":"----------------------------------------"}]}]}
```

I don't think then we should even check for https://github.com/ceramicnetwork/ceramic-anchor-service/blob/develop/src/ancillary/anchor-request-params-parser.ts#L99 if we never return any error no matter what?

ukstv: We **return** an error. We do not _throw_ it, as it breaks composability. If it returns Left value, then CAR file contains something unexpected.

There is a dozen reasons for an error there inside try-catch block apart just CAR file that is not spec-conformant. It could be an IPLD block of a weird structure, even if it is a pretty valid IPLD object. To maintain composability of codecs, we wrap a successful result or an error in an [Either](https://dev.to/attila_vecerek/the-either-monad-3ooc) monad.

PR #1176, opened 2023-10-24
log model stream id

gvelez17:
# model stream id should be logged - #MET-1607

## Description

The log data has been missing the model info. Upon examining a sample CARfile it seems that the genesisfields.header.model is an object, and we are trying to directly log it as a string.  we need to extract the string.

## How Has This Been Tested?

- [x] added tests with a sample carfile that contains a model
- [x] also included checks in tests without a model

## Definition of Done

## References:


src/ancillary/__tests__/anchor-request-params-parser.test.ts
```diff
@@ -125,6 +136,16 @@ describe('AnchoRequestParamsParser', () => {
 
     const params: RequestAnchorParams = (validation as Right<RequestAnchorParams>).right
     expect(params.cacaoDomain).toEqual(FAKE_CACAO)
+    expect((params as RequestAnchorParamsV2).genesisFields.model?.toString() ?? '').toEqual('')
```
stbrody, line 139:
```suggestion
    expect((params as RequestAnchorParamsV2).genesisFields.model).toBeUndefined()
```

stbrody (approved):
lgtm mod one small suggested test cleanup

src/services/request-service.ts
```diff
@@ -109,7 +109,7 @@ export class RequestService {
       did,
       schema: genesisFields?.schema,
       family: genesisFields?.family,
-      model: genesisFields?.model,
+      model: genesisFields?.model?.toString() ?? '',
```
stbrody, line 112: actually wait, I'm not sure this will actually log the right thing.  Have you tested this with a real MID stream's genesis commit?  I think you need to do `StreamID.fromBytes(genesisFields.model).toString()`

stbrody: I'm not 100% positive though, so if you tested it with just calling toString on the Uint8Array and it printed the proper model streamid string we expect, then it's fine.

src/ancillary/__tests__/anchor-request-params-parser.test.ts
gvelez17: well - the way i wrote the code it will be '', not undefined.  I did that because we are logging it and it needs to be a string.

gvelez17: the reason i put this extra cruft in the test was to exercise the similar code in constructing the log data.  but, I agree, it is not really testing the code rather than reproducing it!  I will change the test, but i did  need it this way to make sure it won't break since i didn't have a good test to actually exercise the whole request

src/services/request-service.ts
gvelez17: No, so the thing is, because we decode it actually model is an object.

so, the structure of 'model' depends on whether it was run thru decode or not

```


carFile.get(genesisJWS.link).header.model

Uint8Array(40)


decode(IpfsGenesis, carFile.get(genesisJWS.link)).header.model

{ _tag: Symbol(@ceramicnetwork/streamid/StreamID),

_type: 2,

_cid: CID }


```

and in our case, it is using decode

https://github.com/ceramicnetwork/ceramic-anchor-service/blob/057158e8ab0c735a4a7061dbcca82eaa3bb6e19b/src/services/metadata-service.ts#L98

```


const genesis = decode(IpfsGenesis, genesisRecord)
return genesis.header
```

gvelez17: @stbrody fyi ^^ so i commited your change, will merge when it completes

stbrody: Ah I see, we use the codec to deserialize the genesis commit, so it's already a StreamID instance by the time we get here.  Sounds good.

PR #1180, opened 2023-12-08
feat: add pubsub responder window

stephhuynh18:
We have gotten into a scenario where we were receiving anchor status requests for really old streams/cid combos. In light of this as a safeguard lets only respond to anchors happening in the last three months. Though this is configurable so can be easily changed.

### DIFF
--- a/src/services/__tests__/ipfs-service.test.ts
+++ b/src/services/__tests__/ipfs-service.test.ts
@@ -307,6 +307,53 @@ describe('pubsub', () => {
     expect(receivedTip).toEqual(anchorCid.toString())
   })
 
+  test('Will not respond to query message about stream with an anchor if it is too old', async () => {
+    const pubsubMessage = { typ: 1, id: '1', stream: randomStreamID() }
+
+    const ipfsQueueService = injector.resolve('ipfsQueueService')
+    const queueSendMessageSpy = jest.spyOn(ipfsQueueService, 'sendMessage')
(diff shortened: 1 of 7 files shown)


src/services/__tests__/ipfs-service.test.ts
```diff
@@ -307,6 +307,53 @@ describe('pubsub', () => {
...
+    const pubsubSubscribeSpy = jest.spyOn(mockIpfsClient.pubsub, 'subscribe')
+    pubsubSubscribeSpy.mockImplementationOnce(
+      (topic: string, onMessage: (message: Message) => void) => {
+        expect(topic).toEqual('/faux')
+        onMessage(asIpfsMessage(serialize(pubsubMessage)))
+        return Promise.resolve()
+      }
+    )
+    // @ts-ignore
+    const beforeWindow = new Date(Date.now() - service.pubsubResponderWindowMs - 1000)
```
3benbox, line 325: Small edge case for the test running at a month end.

I think this should be a larger delta than 1000 if we're assuming 31 days in a month

https://github.com/ceramicnetwork/ceramic-anchor-service/pull/1180/files#diff-a6c2c171d98ab607046555ff5f877b97c01087dc716ce801bcb4a0f0f700e524R39

3benbox: never mind, I was reading this wrong.

LGTM

PR #1181, opened 2023-12-08
fix: do not move requests back to pending if error occured during anchoring process is using queue batches

stephhuynh18:
Before: scheduler would take pending anchors and batch them for workers. So if a batch failed workers would move then back to pending so the scheduler would put them in a new batch

now: batches are picked up from a queue, if the batch fails it is automatically put back on the queue and will be picked up by a worker again. No status changes needed

This change should not affect any other modes, and any modes using the older scheduler.


smrz2001 (approved):
🚀

PR #1183, opened 2024-02-20
fix: Return not found error if trying to get status for request that does exist

stephhuynh18:


src/controllers/request-controller.ts
```diff
@@ -82,6 +82,13 @@ export class RequestController {
       const response = await this.requestService.getStatusForCid(cid)
       return res.status(StatusCodes.OK).json(response)
     } catch (err: any) {
+      if (err instanceof RequestDoesNotExistError) {
```
gvelez17, line 85: methinks you also have to make change here?  https://github.com/ceramicnetwork/ceramic-anchor-service/blob/42aad6c012c5033b0f8795d8f5fd241fb06ca721/src/services/request-service.ts#L47

gvelez17: oh nm you did

gvelez17 (approved):
yes this exactly why thank you

PR #1187, opened 2024-03-09
feat: exit with error when an anchor worker fails to anchor a batch

stephhuynh18:
Before: retry 3 times before erroring out
After: no retries, better for debugging

Tested:
tested manually
altered unit test


smrz2001 (approved):
🥳 🚀

PR #1189, opened 2024-03-26
feat: cas scaling

smrz2001:
This PR makes two main changes:
- Puts IPFS store operations behind a new feature flag (`CAS_USE_IPFS_STORAGE`).
- Detects whether a batch queue message was received without request IDs and, if so, pulls the batch from S3. This avoids a large batch from running into the 256KB SQS message size limit.

These changes are backwards compatible. Once ready, they should be merged before the changes from [this](https://github.com/ceramicnetwork/go-cas/pull/38) PR for updating the Scheduler to generate/store larger batches.

### DIFF
--- a/src/repositories/anchor-repository.ts
+++ b/src/repositories/anchor-repository.ts
@@ -6,6 +6,7 @@ import { parseCountResult } from './parse-count-result.util.js'
 import { decode } from 'codeco'
 
 const TABLE_NAME = 'anchor'
+const chunkSize = 10000
 
 export class AnchorRepository implements IAnchorRepository {
   static inject = ['dbConnection'] as const
@@ -32,11 +33,8 @@ export class AnchorRepository implements IAnchorRepository {
    * @returns A promise that resolve to the number of anchors created
    */
   async createAnchors(anchors: Array<FreshAnchor>): Promise<number> {
-    const result: any = await this.table
-      .insert(anchors.map((anchor) => FreshAnchor.encode(anchor)))
-      .onConflict('requestId')
-      .ignore()
-    return parseCountResult(result.rowCount)
+    const result = await this.connection.batchInsert(TABLE_NAME, anchors, chunkSize)
+    return parseCountResult(result.length)
   }
 
   /**
--- a/src/repositories/anchor-repository.ts
+++ b/src/repositories/anchor-repository.ts
@@ -33,7 +33,7 @@ export class AnchorRepository implements IAnchorRepository {
    * @returns A promise that resolve to the number of anchors created
    */
   async createAnchors(anchors: Array<FreshAnchor>): Promise<number> {
-    const result = await this.connection.batchInsert(TABLE_NAME, anchors, chunkSize).catch(function (error) {
+    const result = await this.connection.batchInsert(TABLE_NAME, anchors.map((anchor) => FreshAnchor.encode(anchor)), chunkSize).catch(function (error) {
       console.error(error)
     }) ?? []
     return parseCountResult(result.length)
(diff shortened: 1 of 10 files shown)


stbrody:
Can we split `CAS_USE_IPFS` into two env vars?  One for publishing to pubsub and another for storing the commits themselves?  I think it would be good to turn those off one at a time to help us detect if there are issues.

smrz2001:
> Can we split `CAS_USE_IPFS` into two env vars? One for publishing to pubsub and another for storing the commits themselves? I think it would be good to turn those off one at a time to help us detect if there are issues along the way

I didn't think we'd want to apply the load of either publishing or storing commits to IPFS. I feel like that's just asking for (now gigantic) failed batches due to IPFS issues. Do we really care to spend time figuring out how IPFS does under either type of load? IMO, cutting IPFS out of the picture will be essential for scaling CAS.

stbrody:
I definitely agree that the goal is to stop publishing to ipfs entirely.  I'm just saying let's wind down our use of ipfs in 2 phases.  First let's stop publishing notifications of the anchor commits to pubsub, while continuing to still create them in ipfs so that they are available to bitswap.  Then if we see no problems with that, then we can stop creating the anchor commits against the CAS IPFS entirely.

smrz2001:
> I definitely agree that the goal is to stop publishing to ipfs entirely. I'm just saying let's wind down our use of ipfs in 2 phases. First let's stop publishing notifications of the anchor commits to pubsub, while continuing to still create them in ipfs so that they are available to bitswap. Then if we see no problems with that, then we can stop creating the anchor commits against the CAS IPFS entirely.

I'll add the second flag though I'm quite certain we'll be turning it off right about the time we start benchmarking CAS with 1M sized batches 😉

I'd also posit a different way of looking at this. Our decision to keep IPFS a part of the CAS architecture shouldn't be based on whether or not IPFS has issues during testing, but on whether it's more or less likely to have issues with larger batches. I'm certain that it is much more likely to run into issues with larger batches that we'll then have to mitigate (e.g. writing intermediate Merkle-tree nodes might be be too much too fast and need additional pacing, etc.)

In other words, even if it sails through all of our benchmarking, I would _still_ recommend turning IPFS off completely for production. Whether it works now under high load (I'm skeptical) doesn't mean that it will remain stable in production even under the same type of load.

stbrody:
Sorry, when talked about making decisions based on whether or not there are issues, I wasn't saying that we shouldn't remove IPFS unless it presents performance issues.  I want to remove IPFS ASAP.  I was saying that I'm concerned that once we remove it, we'll have issues in the field from our users where they aren't getting anchor commits.  I'm concerned about any lingering bugs in the anchor polling system delivering anchor commits via CAR files, because up until now we've always had pubsub and bitswap as a backup, so if that isn't totally working we might not ever have noticed.  I'm also worried about users in the wild running old versions of js-ceramic that don't even have the new polling and CAR file features, and might not upgrade or do anything until suddenly their writes start failing with CACAO errors due to missing anchor commits once we turn off publishing of anchors to IPFS.

So *that* is the reason I want to do this in two phases, so that we have time for users in the wild to report issues to us if they stop getting anchor commits, and for us to have a clearer picture where things are breaking down if that happens.

smrz2001:
Yeah, that makes sense, thanks for clarifying. If we have to choose, we can also always setup a separate CAS cluster for V' partners that doesn't use IPFS at all, and leave current CAS with IPFS.

src/repositories/anchor-repository.ts
```diff
@@ -55,6 +55,7 @@ export class AnchorRepository implements IAnchorRepository {
   async findByRequestId(requestId: string): Promise<StoredAnchor | null> {
     const row = await this.table.where({ requestId: requestId }).first()
     if (!row) return null
+    console.log("IM HERE with ", row, requestId)
```
smrz2001, line 58: Should we take this log out, @JulissaDantes? Might flood the logs at high throughput.

JulissaDantes, line 58:
```suggestion
```

JulissaDantes: You are completely right. Thank you for pointing this out!

PR #1190, opened 2024-04-04
fix: only create request if it does not exist

stephhuynh18:
problem: we used to merge on conflict when creating requests. This can result in multiple messages set to the scheduler, and misleading log messages

fix: ignore on conflict 

added unit tests


stbrody (approved):
LGTM!

PR #1193, opened 2024-04-17
Clarify Metric Names - #WS2-3181

gvelez17:
# Clarify Metric Names - #WS2-3181

## Description

When exploring metrics in grafana, it is confusing to see similarly named metrics from the controller, scheduler and workers.  Also, some of the existing names were not semantic as to the meaning.

This PR prefixes all controller metrics with 'c_' and clarifies the meanings.  A future PR may prefix the scheduler and anchor workers separately, but these are of interest for now.

## How Has This Been Tested?

- [ ] Build and unit tests.  Any name mismatches would be caught in teh build.

## Definition of Done

- [ ] If this is merged some grafanas may also need to be updated

## References:

### DIFF
--- a/src/ancillary/anchor-request-params-parser.ts
+++ b/src/ancillary/anchor-request-params-parser.ts
@@ -139,13 +139,13 @@ export class AnchorRequestParamsParser {
   parse(req: ExpReq): Validation<RequestAnchorParams> {
     if (req.get('Content-Type') !== 'application/vnd.ipld.car') {
       // Legacy requests
-      Metrics.count(METRIC_NAMES.LEGACY_REQUESTED, 1)
+      Metrics.count(METRIC_NAMES.C_LEGACY_REQUESTED, 1)
       return validate(RequestAnchorParamsV1, req.body)
     } else {
       // Next version of anchor requests, using the CAR file format
       // TODO: CDB-2212 Store the car file somewhere for future reference/validation of signatures
       // (as it also includes the tip commit and optionally CACAO for the tip commit)
-      Metrics.count(METRIC_NAMES.CAR_REQUESTED, 1)
+      Metrics.count(METRIC_NAMES.C_CAR_REQUESTED, 1)
       return validate(this.v2decoder, req.body)
     }
   }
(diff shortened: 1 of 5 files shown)


src/ancillary/anchor-request-params-parser.ts
```diff
@@ -139,13 +139,13 @@ export class AnchorRequestParamsParser {
   parse(req: ExpReq): Validation<RequestAnchorParams> {
     if (req.get('Content-Type') !== 'application/vnd.ipld.car') {
       // Legacy requests
-      Metrics.count(METRIC_NAMES.LEGACY_REQUESTED, 1)
+      Metrics.count(METRIC_NAMES.C_LEGACY_REQUESTED, 1)
```
3benbox, line 142: What does the `C_` prefix indicate?

3benbox (commented):
Just a question about the prefix.

Also, do these inline comments get added to the metric description?

samika98 (approved):
LGTM! 
Have the same question as Ben, what does the c-prefix mean?

src/ancillary/anchor-request-params-parser.ts
gvelez17: just controller - its to group together the controller ones so its not confusing with the service workers or scheduler

we could spell it out but i feel like super long names are not that efficient - if prefered could spell it out or put ctrl_ instead of c_

gvelez17: anyway, if it was confusing once it will be again, changed it to 'ctrl_' @3benbox  ^

gvelez17:
@ukstv @stephhuynh18 we're changing some metric names in cas, i dont think you rely on any of these but FYI just in case.

src/ancillary/anchor-request-params-parser.ts
stbrody: maybe RC_ for RequestController?

gvelez17:
> Just a question about the prefix.
> 
> Also, do these inline comments get added to the metric description?

so, right now I was not setting the metric description :-| 
https://github.com/ceramicnetwork/observability/blob/master/packages/service-metrics/src/service-metrics.ts#L287

the only thing is if i start setting it, the behavior is not well defined, if the prometheus node already has that metric name without description, it might not add it.  I'll make a separate ticket to add descriptions, and yes this would make them much more usable!

gvelez17:
huh i just realized these will already be prefixed cas_server, because it uses the 'mode'

```
Metrics.start(
        this.config.metrics.collectorHost,
        'cas_' + this.mode,
        ```
        
        https://github.com/ceramicnetwork/ceramic-anchor-service/blob/d370fd10fa573b7237b435fb6a6ab7fd5e781198/src/app.ts#L119
        
        I think in some cases the 'mode' was blank

PR #1194, opened 2024-04-19
fix: move db connection outside handler, too many open files

3benbox:
During load testing on CAS, we're getting error "EMFILE: too many files open".

### DIFF
--- a/auth/src/authorizers/default.ts
+++ b/auth/src/authorizers/default.ts
@@ -17,6 +17,8 @@ const allowRegisteredDIDSchema = Joi.object({
   nonce: nonceValidation.required()
 })
 
+const dynamodbClient = new DynamoDB(false)
+
 export const handler = async (event: APIGatewayRequestAuthorizerEvent, context, callback) => {
   console.log(event)
   console.log(context)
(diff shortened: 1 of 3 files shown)


smrz2001 (approved):
🚀

auth/src/authorizers/default.ts
```diff
@@ -99,9 +101,7 @@ async function allowRegisteredDID(event: APIGatewayRequestAuthorizerEvent, callb
       if (error) {
         console.error(error.details)
       } else {
-        const createTableIfNotExists = false
-        const db = new DynamoDB(createTableIfNotExists)
-        const data = await db.addNonce(did, nonce)
+        const data = await dynamodbClient.addNonce(did, nonce)
```
gvelez17, line 104: good catch!  if its long running tho and the handle closes will we need to reopen it?  Could we possibly test this branch in the keramik before merge to develop?

gvelez17:
Great catch!  Have we tested it?

3benbox:
> Great catch! Have we tested it?

It's running on dev now.

auth/src/authorizers/default.ts
3benbox: I hope the sdk client would restart the connection 🙏 

The current problem is the lamba is running too long and we think it's opening too many connections.

gvelez17:
um - won't the next time we merge to develop automatically deploy develop branch to dev-cas - i think if its already running on dev cas we should def merge this?

smrz2001:
> um - won't the next time we merge to develop automatically deploy develop branch to dev-cas - i think if its already running on dev cas we should def merge this?

We never had CD for CAS Auth unfortunately. Whenever we merge this, the CD will run but that just deploys regular CAS code, not Auth code. We'll have to manually deploy the Lambda to all the envs.

gvelez17:
per Mohson we are not affecting the anchoring in this pr, force-merging

PR #1196, opened 2024-04-19
chore: add logs for post anchoring steps

smrz2001:


samika98 (approved):
Do we want to add merkle root cid in these logs?

PR #1197, opened 2024-04-22
feat: do not write to ipfs

smrz2001:
This PR puts IPFS writes behind a feature flag. We can add more things for operationalizing CAS to this PR, if needed (e.g. more metrics).

### DIFF
--- a/src/__tests__/ceramic_integration.test.ts
+++ b/src/__tests__/ceramic_integration.test.ts
@@ -265,15 +265,24 @@ describe('Ceramic Integration Test', () => {
   beforeAll(async () => {
     ipfsApiPort1 = await getPort()
     ipfsApiPort2 = await getPort()
-    ;[ipfs1, ipfs2, ipfs3, ipfs4] = await Promise.all([
-      createIPFS(ipfsApiPort1),
-      createIPFS(ipfsApiPort2),
-      createIPFS(),
-      createIPFS(),
-    ])
+    if (process.env['CAS_USE_IPFS_STORAGE']) {
+      ;[ipfs1, ipfs2, ipfs3, ipfs4] = await Promise.all([
+        createIPFS(ipfsApiPort1),
+        createIPFS(ipfsApiPort2),
+        createIPFS(),
+        createIPFS(),
+      ])
+    } else {
+      ;[ipfs3, ipfs4] = await Promise.all([
+        createIPFS(ipfsApiPort1),
+        createIPFS(ipfsApiPort2),
+        createIPFS(),
+        createIPFS(),
+      ])
+    }
 
     // Now make sure all ipfs nodes are connected to all other ipfs nodes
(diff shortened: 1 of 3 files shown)


stbrody (approved):
why am I having deja vu?  Didn't we do this already?

stbrody (commented):
also lets make sure to set that ENV var on all our environments before merging this, or else we risk changing behavior immediately.

Should we maybe invert the condition so the default behavior stays the same as it is now, and it requires an active config change to change the behavior and stop using ipfs?

smrz2001:
> why am I having deja vu?  Didn't we do this already?

Lolol yes, we did, but in the scaling prototype PR, which is on hold. I pulled this change from that PR. 

> also lets make sure to set that ENV var on all our environments before merging this, or else we risk changing behavior immediately.

Good point, will do. 

> Should we maybe invert the condition so the default behavior stays the same as it is now, and it requires an active config change to change the behavior and stop using ipfs?

Honestly, I'd rather add the new variable to the envs than have a double negative haha.

`CAS_DONT_USE_IPFS_STORAGE = false` will break my brain...

https://tenor.com/bDsu7.gif

3benbox:
> > Should we maybe invert the condition so the default behavior stays the same as it is now, and it requires an active config change to change the behavior and stop using ipfs?
> 
> Honestly, I'd rather add the new variable to the envs than have a double negative haha.
> 
> `CAS_DONT_USE_IPFS_STORAGE = false` will break my brain...

How about a metric when are successfully loaded on IPFS, in addition to the current failure metric.

That way we can track that ipfs loads are still happening where we expect them to be.

3benbox (approved):
lgtm!

smrz2001:
> How about a metric when are successfully loaded on IPFS, in addition to the current failure metric.
> 
> That way we can track that ipfs loads are still happening where we expect them to be.

Yeah, I don't mind that. I can add new metrics for IPFS record and CAR file storage.

smrz2001:
FYI, @stbrody @ukstv @stephhuynh18, I haven't added the new env var to any of the environments. Since the deadline for deprecating CAS IPFS is coming up next week anyway, IMO we can leave IPFS disabled by default.

If the PR looks good, we can merge to Dev/QA. Then Monday we can merge to Clay, watch a couple of days, then plan to merge to Prod. If something blows up, we can add the env var and re-enable IPFS quickly.

stbrody:
Sounds good, definitely let's get this merged ASAP so it can start burning in and we can see if any problems show up in our tests.

I do still think we should probably spend some time looking into the merkle car errors on prod before we deploy to prod: https://linear.app/3boxlabs/issue/WS1-1608/no-merkle-car-found-for-anchor-errors-are-common-on-prod-cas.  CC @ukstv @christianlavoie

stephhuynh18 (approved):

PR #1198, opened 2024-04-22
add capability to set an instance identifier which labels all metrics

gvelez17:
# set instance identifier - #WS2-3187
## Description

The current CAS metrics are being sent from the 6 different api nodes with the same name and labels, resulting in inaccurate metrics, stepping on totals, odd reset patterns.  Each instance should have a distinct label.  This PR uses the task id as the label and uses the new observability functionality of Instance Identifier to automatically label all metrics (see https://github.com/ceramicnetwork/observability/pull/29)

## How Has This Been Tested?

- [ x] observability package new unit tests
- [ ] unit tests in this pr

## Definition of Done

- [x ] The work addresses the description and outcomes in the issue
- [ x] I have added relevant tests for new or updated functionality

## References:

### DIFF
--- a/config/env/dev.json
+++ b/config/env/dev.json
@@ -56,7 +56,7 @@
     "traceRatio": "@@METRICS_TRACE_RATIO",
     "exportIntervalMillis": "@@METRICS_EXPORT_INTERVAL_MS",
     "exportTimeoutMillis": "@@METRICS_EXPORT_TIMEOUT_MS",
-    "instanceIdentifier": "@@ECS_CONTAINER_METADATA_URI",
+    "instanceIdentifier": "@@INSTANCE_IDENTIFIER",
     "prometheusPort": "@@METRICS_PORT"
   },
   "db": {
--- a/config/env/dev.json
+++ b/config/env/dev.json
@@ -56,6 +56,7 @@
     "traceRatio": "@@METRICS_TRACE_RATIO",
     "exportIntervalMillis": "@@METRICS_EXPORT_INTERVAL_MS",
     "exportTimeoutMillis": "@@METRICS_EXPORT_TIMEOUT_MS",
+    "instanceIdentifier": "@@ECS_CONTAINER_METADATA_URI",
     "prometheusPort": "@@METRICS_PORT"
   },
   "db": {
--- a/src/app.ts
+++ b/src/app.ts
@@ -135,8 +135,11 @@ export class CeramicAnchorApp {
 
          // in this case, return only the task id following the last /
          const match = this.config.metrics.instanceIdentifier.match(/\/([^\/]*)$/);
-         const instanceId: string = match ? match[1] : this.config.metrics.instanceIdentifier
-         Metrics.setInstanceIdentifier(instanceId)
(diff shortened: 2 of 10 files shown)


config/env/dev.json
```diff
@@ -56,6 +56,7 @@
     "traceRatio": "@@METRICS_TRACE_RATIO",
     "exportIntervalMillis": "@@METRICS_EXPORT_INTERVAL_MS",
     "exportTimeoutMillis": "@@METRICS_EXPORT_TIMEOUT_MS",
+    "instanceIdentifier": "@@ECS_CONTAINER_METADATA_URI",
```
3benbox, line 59: I'm not familiar with the `@@` syntax. This will set the value to the corresponding env var?

3benbox (commented):
Just a question about how the value gets set.

config/env/dev.json
gvelez17: yes that's correct - so if present, it is going to set it from ECS_CONTAINER_METADATA_URI which is present on the api server instances.  It will not get set otherwise and won't be used as a label if its not present

3benbox (approved):
lgtm

config/env/dev.json
smrz2001, line 59: It would be nice to inject this through the Dockerfile via an ENTRYPOINT script and call it `INSTANCE_ID` or something like that.

We do have AWS specific stuff in the env vars and code but parsing a very specific ECS env var directly in the code seems undesirable.

`entrypoint.sh`:
```
#!/bin/bash

uri=$ECS_CONTAINER_METADATA_URI

# Parse the last portion of the URI
value=${uri##*/}

# Remove everything after the hyphen to get the task ID
value=${value%%-*}

export INSTANCE_ID=$value

exec "$@"
```

`Dockerfile`:
```
.
.
.
WORKDIR /

COPY entrypoint.sh .
RUN chmod +x ./entrypoint.sh

COPY runner.sh .

ENTRYPOINT ["./entrypoint.sh"]
CMD [ "./runner.sh" ]
```

src/app.ts
```diff
@@ -129,6 +129,18 @@ export class CeramicAnchorApp {
       )
       Metrics.count('HELLO', 1)
       logger.imp('Metrics exporter started')
+      if (this.config.metrics.instanceIdentifier) {
+         // In the API server, the instanceIdentifer is set from ECS_CONTAINER_METADATA_URI
```
smrz2001, line 133: If done using the `entrypoint.sh` script above, we won't need to parse this in the code here.

config/env/dev.json
3benbox: > We do have AWS specific stuff in the env vars and code but parsing a very specific ECS env var directly in the code seems undesirable.

Oh yeah, good call!

src/app.ts
gvelez17: right!  the logic isn't quite the same - not sure if it matters, i was defaulting to the full uri if there wasn't a task id at the end - but i agree yours is better, since it moves it to the deployment and therefore we can make it very specific to what we expect!

PR #1200, opened 2024-05-06
chore: use node v20

smrz2001:


gvelez17 (approved):
lets try it on dev?  if it builds and runs, likely it is ok?

ukstv (approved):
<flying colours emoji>

PR #1201, opened 2024-05-07
Fix: retry db transaction if we get a serialization error when persisting anchors

stephhuynh18:
Fixes error "could not serialize access due to concurrent update". This is the only place that does not catch this error and retry

### DIFF
--- a/src/services/anchor-service.ts
+++ b/src/services/anchor-service.ts
@@ -23,6 +23,7 @@ import { v4 as uuidv4 } from 'uuid'
 import type { Knex } from 'knex'
 import type { IIpfsService } from './ipfs-service.type.js'
 import type { IAnchorRepository } from '../repositories/anchor-repository.type.js'
+import { REPEATED_READ_SERIALIZATION_ERROR } from '../repositories/repository-types.js'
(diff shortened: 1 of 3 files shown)


gvelez17 (approved):
LGTM

smrz2001 (approved):
🥳

src/services/anchor-service.ts
```diff
@@ -546,40 +547,46 @@ export class AnchorService {
...
-      completed.publishStats(METRIC_NAMES.CREATED_SUCCESS_MS)
-      Metrics.count(METRIC_NAMES.ACCEPTED_REQUESTS, acceptedRequests.length)
-
-      return persistedAnchorsCount
-    } catch (err) {
-      await trx.rollback()
-      throw err
-    }
+      .catch(async (err) => {
+        if (err?.code === REPEATED_READ_SERIALIZATION_ERROR) {
```
stbrody, line 582: should we still log the error here?

stephhuynh18: I can add a log sure

yes the rollback is handled automatically

PR #1202, opened 2024-05-07
fix: add tests and log when there is an error when persisting anchors

stephhuynh18:
i forgot the test when creating the fix (https://github.com/ceramicnetwork/ceramic-anchor-service/pull/1201)

### DIFF
--- a/src/services/__tests__/anchor-service.test.ts
+++ b/src/services/__tests__/anchor-service.test.ts
@@ -378,6 +378,23 @@ describe('anchor service', () => {
     expect(anchors.length).toEqual(numRequests)
   })
 
+  test('Retry persisting anchors on serializable erorr', async () => {
+    const requests = await fake.multipleRequests(4)
+
+    const [candidates] = await anchorService._findCandidates(requests, 0)
+    const merkleTree = await anchorService._buildMerkleTree(candidates)
+    const ipfsProofCid = await ipfsService.storeRecord({})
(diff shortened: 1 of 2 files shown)


src/services/__tests__/anchor-service.test.ts
```diff
@@ -378,6 +378,23 @@ describe('anchor service', () => {
...
 
+  test('Retry persisting anchors on serializable erorr', async () => {
+    const requests = await fake.multipleRequests(4)
+
+    const [candidates] = await anchorService._findCandidates(requests, 0)
+    const merkleTree = await anchorService._buildMerkleTree(candidates)
+    const ipfsProofCid = await ipfsService.storeRecord({})
+    const anchors = await anchorService._createAnchorCommits(ipfsProofCid, merkleTree)
+
+    await Promise.all([
```
stbrody, line 389: I guess we're relying on a race condition here to *sometimes* trigger the serialization error we are catching and retrying on?  I would prefer if we inject the failure explicitly so it's deterministic, but also don't feel super strongly.  Can we at least add a comment explaining what we're doing here and why?

stbrody (approved):
lgtm

src/services/__tests__/anchor-service.test.ts
stephhuynh18: made the test more deterministic and added comments

PR #1205, opened 2024-05-08
fix: exit runner scripts explicitly once completed

smrz2001:
Explicitly exit the runner scripts once they are complete.


3benbox (approved):
LGTM

PR #1207, opened 2024-05-09
feat: always install datadog package in images

3benbox:
This package won't do anything unless initialized and should be safe for all images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,6 @@ EXPOSE 8081
 
 CMD npm run start
 
-FROM base as tracing
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,8 +20,6 @@ EXPOSE 8081
 
 CMD npm run start
 
-FROM base as tracing
-
 RUN npm install dd-trace --save
```
smrz2001, line 23: Should we move the install to before the npm run start? Maybe to line 14 after `npm run ci`?

smrz2001 (approved):
🚀

PR #1209, opened 2024-05-14
fix: simplify runner scripts

smrz2001:


Dockerfile
```diff
@@ -38,18 +35,12 @@ ENV AWS_ECS_FAMILY=${AWS_ECS_FAMILY}
...
-
-WORKDIR /
-
-COPY entrypoint.sh .
-RUN chmod +x ./entrypoint.sh
+WORKDIR /cas
 
-COPY runner.sh .
+COPY runner ./runner
+COPY runner.sh ./
```
3benbox, line 41: Isn't this copying the files to /cas?
Should we copy to root
```
COPY runner.sh /runner.sh
etc.
```

3benbox: ok, looks like the paths are relative and builds/runs fine for me!

3benbox (approved):
LGTM

PR #1211, opened 2024-05-15
upgrade observability package for better error logging

gvelez17:
# Upgrade observability package - better error logging

## Description

This just provides better error logging when metrics fails to start because of invalid parameters


gvelez17:
weird a rebase shouldn't have duplicated the commits, i dunno why that happened

PR #1212, opened 2024-05-17
feat: read replica db for findByCid

samika98:
# [Read Replica] - #[https://linear.app/3boxlabs/issue/WS2-3238/create-a-separate-read-replica-to-improve-cas-performace]

## Description

We noticed Knex connection pools getting exhausted during high load after some time. In order to scale CAS we have to introduce a separate read replica to handle the read calls. This will distribute the load between the replicas giving us better throughput

This PR has env vars defined to access the CAS-API : https://github.com/3box/ceramic-infra/pull/798/files#diff-7a1b7f1deef12acc800844bd1061eecf14fc08ded251a2363f32325a3210fcea
## How Has This Been Tested?
TODO : Saw decreased latency monitored via data dog for read queries. Earlier read queries would take ~3-4 minutes under high load. We observe time in order of milli seconds for reads with this change 

## References:
Update ts for typed-inject compatibility : https://github.com/microsoft/TypeScript/issues/37400. For some reason this bug was still present on ts 5.0.4 but not on the latest version which is 5.4.5
https://www.npmjs.com/package/typed-inject


christianlavoie (commented):
How does this affect existing metrics? Are there existing db query centric metrics and how are they affected with the two connection pools?

samika98:
> How does this affect existing metrics? Are there existing db query centric metrics and how are they affected with the two connection pools?
@christianlavoie the only db metric we have currently in CAS is db_serialiazation_errors, this should not be affected by the two connection pools. Other metrics on AWS RDS should get affected, we would see more reads on the replica and decline of reads in the main db. [See this dashboard for AWS metrics which can change](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#explorer?config=~(timeRange~3600000~model~(metrics~(~(metricName~'CPUUtilization~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Average)~(metricName~'ReadLatency~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Average)~(metricName~'DatabaseConnections~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Sum)~(metricName~'FreeStorageSpace~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Average)~(metricName~'FreeableMemory~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Average)~(metricName~'ReadThroughput~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Average)~(metricName~'ReadIOPS~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Average)~(metricName~'WriteLatency~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Average)~(metricName~'WriteThroughput~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Average)~(metricName~'WriteIOPS~resourceType~'AWS*3a*3aRDS*3a*3aDBInstance~stat~'Average))~labels~(~(key~'AnchorService~value~'true))~widgetOptions~(legend~(position~'bottom)~view~'timeSeries~stacked~false~rowsPerPage~50~widgetsPerRow~2)~period~60~showConfig~true~splitBy~'~region~'us-east-2)~timeZone~'LOCAL))

christianlavoie (commented):
LGTM, but I don't know the codebase enough to feel comfortable approving.

src/services/request-service.ts
```diff
@@ -40,19 +42,29 @@ export class RequestService {
...
+   * @param cid The CID of the request.
+   * @returns The request.
+   */
   async getStatusForCid(cid: CID): Promise<OutputOf<typeof CASResponse> | { error: string }> {
-    const request = await this.requestRepository.findByCid(cid)
+    let request = await this.replicationRequestRepository.findByCid(cid)
     if (!request) {
-      throw new RequestDoesNotExistError(cid)
+      request = await this.requestRepository.findByCid(cid)
+      logger.debug(`Request not found in replica db for ${cid}, fetching from main_db`)
```
stephhuynh18, line 62: Nitpick: probably should go before line 61

src/services/request-service.ts
```diff
@@ -40,19 +42,29 @@ export class RequestService {
...
+      Metrics.count(METRIC_NAMES.REPLICA_DB_REQUEST_NOT_FOUND, 1)
```
stephhuynh18, line 63: Do we want to move this above line 61 (if line 61 errors, the metric won't be counted)

stephhuynh18: Same for below

src/services/request-service.ts
stephhuynh18: Same for below

src/db-connection.ts
```diff
@@ -69,6 +69,38 @@ export async function createDbConnection(dbConfig: Db = config.db): Promise<Knex
   return connection
 }
 
+export async function createReplicaDbConnection(
```
stephhuynh18, line 72: instead of having a separate function can we reuse the one above , I like the explicit key value pairs

stephhuynh18: Also does this also need to do migrations (in case of future)

samika98: I like having two separate functions. So that if in the future we want to disable a read replica, it would be easier to add that as a config, and add a check in app.ts. I like having a function do just one thing. 
No we don't need any migrations, rds takes care of that

src/services/request-service.ts
samika98: Thanks for catching this!

stephhuynh18 (approved):

PR #1215, opened 2024-05-20
feat: cas scaling ph 2 (s3 batching)

smrz2001:
Allow workers to be able to pull anchor batches from S3, with fallback to existing logic for extracting batches from SQS messages.

This needs to be merged before S3 batching in the Scheduler is merged, since this code is backward-compatible and capable of handling both types of batching strategies.


stephhuynh18 (approved):
🎈

PR #1216, opened 2024-05-20
feat: remove separate runner image

smrz2001:
And limit anchor contract test workflow execution to actual changes to the contract...


3benbox (approved):
lgtm

PR #1218, opened 2024-05-21
feat: batch db calls when loading candidates

stephhuynh18:
Currently we loop through each request and retrieve the metadata for it, and check if there is an anchor for it. Instead of doing this per request, we can do one batch call to do it before processing.


smrz2001 (approved):
🚀

PR #1219, opened 2024-05-29
feat: Do not build TreeMetadata into anchor merkle trees

stbrody:
This is the first step of removing request CAR files from the CAS entirely.  With this PR we no longer ever read the metadata out of the metadata service that is populated via the request CAR file.  The next PR will stop us from writing that metadata to the database at all.


src/services/anchor-service.ts
```diff
@@ -647,19 +645,11 @@ export class AnchorService {
...
-      })
 
     for (const request of requests) {
       const streamId = StreamID.fromString(request.streamId)
-      const metadata = metadataByStreamId[request.streamId]
-      if (metadata) {
-        const candidate = new Candidate(streamId, request, metadata.metadata)
-        candidates.push(candidate)
-      }
+      const candidate = new Candidate(streamId, request, null as unknown as ICandidateMetadata)
```
stbrody, line 651: it's annoying that I can't just easily change `Candidate` to not accept metadata at all, or to skip sorting the leaf nodes entirely.  But that code lives in the anchor-utils package in the js-ceramic repo now, and the CAS repo is still using extremely old versions of the js-ceramic packages.  So rather than go through the headache of first upgrading all the js-ceramic packages in the CAS repo, and then updating the anchor-utils package with new versions of the merkle tree builder apis, and then coming back to the CAS to update it to use the updated anchor-utils, I took the shortcut of leaving all the code for building metadata and sorting leaf nodes in place in the underlying code for building merkle trees, I just don't pass a function for building metadata and provide a very simplified leaf sorter.

ukstv: That's an immensely helpful comment!

Any plans to change js-ceramic counterpart?

src/merkle/merkle-car-factory.ts
```diff
@@ -2,22 +2,32 @@ import type { DiagnosticsLogger } from '@ceramicnetwork/common'
...
   type Node,
   type PathDirection,
   type CIDHolder,
-  type TreeMetadata,
+  type TreeMetadata, CompareFunction,
 } from '@ceramicnetwork/anchor-utils'
-import { BloomMetadata } from './bloom-metadata.js'
 import type { Candidate } from '../services/candidate.js'
 import type { CID } from 'multiformats/cid'
+import {ICandidate} from "@ceramicnetwork/anchor-utils/lib/merkle/candidate.type.js";
```
ukstv, line 15: That probably will break in production.

ukstv (changes requested):
Hacky in a gooood way ^_^ LGTM, with two nitpicks.

I suspect, that could sound annoying: please, prettier :D

Also, I am cautious about the plain JS import (see a separate comment). This might break Node.js mode resolution algo.

src/merkle/merkle-car-factory.ts
ukstv: Huh, I am positively surprised the code passed CI.

Ah, it is `type`! That's why!

ukstv, line 15:
```suggestion
import type { ICandidate } from '@ceramicnetwork/anchor-utils'
```

stbrody: ooh good catch, fixed.

src/services/anchor-service.ts
stbrody: Currently no, not planning to update the types in js-ceramic.  We certainly could get a lot of code cleanup if we do, but in order to use those types here we have to first update the CAS to newer versions of all of the js-ceramic packages.  Might very well be worth doing if we wind up doing much more work on the CAS, but since our current plan is to wrap up CAS scaling work soon and then not touch the CAS anymore until we kill it entirely in favor of self-anchoring, my hope is we can just ignore this code until it goes away.

Happy to revisit that though if this old messy code starts getting in the way of our ability to move fast on any of the projects we need in the future.

stbrody (commented):
Fixed the import.  Agreed about running prettier, but when I tried it it made a LOT of changes on things unrelated to this PR, so I'd like to merge this then do a dedicated PR to just run prettier across the whole repo.

ukstv (approved):

PR #1220, opened 2024-05-29
feat: reuse connections in http client + increase max sockets

samika98:
# [Scale SQS] - #[Issue]

## Description

SQS sendMessages  is taking  >3 minutes. We are increasing the number of socket connections sqs can make and adding a config to allow it to reuse connections.

## References:
https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/node-configuring-maxsockets.html

### DIFF
--- a/src/services/queue/sqs-queue-service.ts
+++ b/src/services/queue/sqs-queue-service.ts
@@ -128,8 +128,8 @@ export class SqsQueueService<TValue extends QueueMessageData>
       requestHandler: new NodeHttpHandler({
         httpAgent: new http.Agent({
           keepAlive: true,
-          maxSockets: 50,
-          maxFreeSockets: 20,
+          maxSockets: 1000,
+          maxFreeSockets: 768,
         }),
       }),
     })
--- a/src/services/queue/sqs-queue-service.ts
+++ b/src/services/queue/sqs-queue-service.ts
@@ -128,8 +128,6 @@ export class SqsQueueService<TValue extends QueueMessageData>
       requestHandler: new NodeHttpHandler({
         httpAgent: new http.Agent({
           keepAlive: true,
-          maxSockets: 1000,
-          maxFreeSockets: 768,
         }),
       }),
     })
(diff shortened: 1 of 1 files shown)


src/services/queue/sqs-queue-service.ts
```diff
@@ -123,6 +125,13 @@ export class SqsQueueService<TValue extends QueueMessageData>
       region: config.queue.awsRegion,
       endpoint: this.sqsQueueUrl,
       logger: awsLogger,
+      requestHandler: new NodeHttpHandler({
+        httpAgent: new http.Agent({
+          keepAlive: true,
+          maxSockets: 50,
+          maxFreeSockets: 20,
```
dav1do, line 132: Why set this one? Are we worried about the default? https://nodejs.org/dist/latest-v4.x/docs/api/http.html#http_agent_maxfreesockets

dav1do (commented):
Looks okay to me. One question and you linked v2 documentation, it's not that different in this case 😄

src/services/queue/sqs-queue-service.ts
samika98: 256 seems a lot, added this to mitigate the extra overhead, if max is 50

dav1do: Yeah, it would basically mean all 50. 256 is a lot less than the other default of infinity (infinitely less even lol). I don’t really mind either way but I’m not sure it’s worth modifying unless we have a symptom like running out of file ulimits or seeing memory pressure. Setting it makes it seem like we expect it to matter 🤷

dav1do (approved):

src/services/queue/sqs-queue-service.ts
dav1do: If we aren’t sure what we want yet and want to experiment a bit, we could have `process.env.someGoodVarName || 50`. Not sure it’s worth exposing as config, but it’s an option

dav1do: Also (sorry to pile on with comments), the title isn’t quite right as 50 should be the default so we’re not increasing it, and 20 free is less than the default. It’s the reuse that’s being modified

samika98: That's problematic, let me bump these up. I think 256 is not enough, let me double it then to 512

smrz2001 (approved):
🥳

src/services/queue/sqs-queue-service.ts
samika98: removing these, let's tart with just the keep alive. Reuse the connections that should reduce the number of dns lookup calls

PR #1221, opened 2024-05-31
Fix/default to maindb if read replica errs

samika98:

## Description

## How Has This Been Tested?

- [ ] Test B

## Definition of Done

## References:

### DIFF
--- a/src/db-connection.ts
+++ b/src/db-connection.ts
@@ -71,7 +71,7 @@ export async function createDbConnection(dbConfig: Db = config.db): Promise<Knex
 
 export async function createReplicaDbConnection(
   replica_db_config: Replicadb = config.replica_db
-): Promise<Knex> {
+): Promise<{ connection: Knex; type: string }> {
   const replicaKnexConfig: Knex.Config = {
     client: replica_db_config.client,
     connection: replica_db_config.connection.connectionString || {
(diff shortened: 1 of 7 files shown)


src/services/request-service.ts
```diff
@@ -79,11 +84,16 @@ export class RequestService {
    * @returns The request.
    */
   async findByCid(cid: CID): Promise<OutputOf<typeof CASResponse> | undefined> {
-    let found = await this.replicationRequestRepository.findByCid(cid)
+    let found
+    try {
+      found = await this.replicationRequestRepository.findByCid(cid)
```
dav1do, line 89: My preference would be to expect the replica to work if it exists, but default the config to main if there is none defined. I think we should expect it to work if it's defined, but it should be optional.

dav1do: In terms of implementation, I'd expect we either set up a RO connection pool, or we just use the main pool and the app still calls `replicationRequestRepository` for all reads because it doesn't require the writer, without needing to know anything about the actual DB used.

This also avoids doing double reads for things that don't actually exist and potentially getting hung trying to talk to a non-exists RO instance.

samika98: yeah, it all around a better design to make the read replication optional. FYI, currently all our reads do not got to the replica. If the configuration for the read replica is not specified or if any of the config params are missing, we default to the maindb. Will add a commit for this and tag you, let me know if this might fix it or there might be something else I missed

src/db-connection.ts
```diff
@@ -71,7 +71,7 @@ export async function createDbConnection(dbConfig: Db = config.db): Promise<Knex
 
 export async function createReplicaDbConnection(
   replica_db_config: Replicadb = config.replica_db
-): Promise<Knex> {
+): Promise<{ connection: Knex; type: string }> {
```
samika98, line 74: @dav1do this change makes the readReplica optional and by introducing a connectionType we have captured the two modes it can work in. 
The type is also useful to avoid misleading logs in the connection failure mode giving devs more insight into what is happening in the background

src/services/request-service.ts
dav1do: I like the changes you made, do we still need this fallback? if it fails now, it seems like it's failing for a reason so there's not sure we need to try another database.

dav1do (approved):
Looks good. One more comment since I'm not sure we need to fallback now but approving.

src/services/request-service.ts
dav1do: Yeah, what I was trying to say is that if the RR doesn't have the request, it's unlikely the main DB does so there's no reason to ask. If for some reason we're far enough behind on replicating (i.e. minutes), we probably don't want to load the main database more and should investigate. Otherwise, we can tolerate the client retrying again in a few seconds and the RR will respond with data.

PR #1222, opened 2024-06-03
fix: log appropriate message when retrying db tx due to serialization error

stephhuynh18:
https://linear.app/3boxlabs/issue/AES-126/anchor-persistence-failures

Errors are showing up due to seralization errors, but they do not indicate that there is a retry mechanism


smrz2001 (approved):
🚀

stbrody (approved):

PR #1223, opened 2024-06-03
feat: Remove all parsing, storing, and handling of stream metadata from genesis commit

stbrody:
Builds on https://github.com/ceramicnetwork/ceramic-anchor-service/pull/1219

Now that we do not build the TreeMetadata object into the anchor merkle tree, we no longer need to extract stream metadata from the genesis commit of the stream being anchored.  This lets us completely remove the MetadataService and all of its usages.

Note that the CAS will still do the work of parsing the CAR file, since it needs the basic information from the root object of the CAR for things like the CID being anchored.  But now we don't use anything in the CAR file other than the root object, the object with the genesis commit is now ignored.

The last PR of this sequence will be changing how js-ceramic requests anchors to no longer send a CAR file at all, and go back to the simple HTTP POST method to provide the CID and streamId being anchored.


stbrody:
Note that this removes the "cacaoDomain" that we currently extract from the CAR file data for the event being anchored.  Not sure if this is used in our metrics dashboards currently, but it means the CAS will now only have access to the DID and ip addressed used to create the request, not the cacao domain from the event being requested itself (but we were going to lose that visibility when switching to self-anchoring and/or delegated anchoring anyway).  FYI @gvelez17 @christianlavoie

ukstv (approved):

PR #1229, opened 2024-06-18
feat: include batch details in anchor results log

smrz2001:


stephhuynh18 (approved):
🐮

src/services/anchor-service.ts
```diff
@@ -50,7 +50,11 @@ type RequestGroups = {
 }
 
 type AnchorSummary = {
+  // id of the batch
+  batchId: string
   // all requests included in this batch
+  requests: number
```
stbrody, line 56: sorry I know this is already merged but there are a lot of fields in this summary that are some sort of count of requests.  Just calling this `requests` I'm worried will be confusing to someone trying to understand what this field represents and how it's different than `anchorCount`, `acceptedRequestsCount`, etc.  Can we rename this to something a bit more descriptive (and add the Count suffix for consistency with the other fields in this object)?

PR #1232, opened 2024-06-20
feat: include js-ceramic and ceramic-one version in request creation metric

stbrody:
When using ceramic-one as the underlying event storage the "ceramicOneVersion" will be reported as something like `"ceramic-one/0.19.0"`.  When using ipfs/kubo, it looks something like `"kubo/0.18.1/"`

This is the CAS-side of the change, the corresponding js-ceramic change is in https://github.com/ceramicnetwork/js-ceramic/pull/3246


src/controllers/request-controller.ts
```diff
@@ -124,7 +125,12 @@ export class RequestController {
 
       // request was newly created
       if (body) {
-        Metrics.count(METRIC_NAMES.CTRL_NEW_ANCHOR_REQUEST, 1, { source: parseOrigin(req) })
+        const v3Params = requestParams as RequestAnchorParamsV3
+        Metrics.count(METRIC_NAMES.CTRL_NEW_ANCHOR_REQUEST, 1, {
+          source: parseOrigin(req),
+          jsCeramicVersion: v3Params.jsCeramicVersion,
```
stbrody, line 131: @gvelez17 is this sufficient to get what we want to track what version people are running based on the requests coming into CAS?

src/__tests__/ceramic_integration.test.ts
```diff
@@ -1,684 +0,0 @@
-import 'reflect-metadata'
```
stbrody, line 1: This test has been flaky for a while but is failing consistently on this PR.  I'm pretty sure it's because I changed the timestamp to be required instead of optional, and we are still using a very old version of js-ceramic in the CAS repo.  Upgrading the CAS to the newest version of js-ceramic is non-trivial and frankly doesn't seem worth it just for test code, now that the CAS doesn't use ceramic in its actual production code.

When I first wrote this test we had very little in the way of test coverage that the CAS was working and that ceramic and the CAS were playing well together.  Nowadays we have a lot more test coverage and have the ceramic-cas integration tested in several other places.  Because of that I think maintaining this test is becoming more trouble than it's worth.

FYI @smrz2001

src/controllers/request-controller.ts
smrz2001: Should we also add this to `getStatusForCid` for status requests?

stbrody: We certainly could, but seems less useful and there is *some* cost in bandwidth to sending this on every request.

stbrody: I figured how many requests are created by which versions was a more useful metric than how much polling is happening from which version

gvelez17: it should be - we don't really want more than 2 version parameters per node, I would think.  these are the key ones.  If something else changes we can try to enforce bumping the minor version of js-ceramic to recognize it

gvelez17 (approved):
LGTM

src/ancillary/anchor-request-params-parser.ts
```diff
@@ -96,7 +98,7 @@ export class AnchorRequestParamsParser {
     if (req.get('Content-Type') !== 'application/vnd.ipld.car') {
       // Legacy requests
       Metrics.count(METRIC_NAMES.CTRL_LEGACY_REQUESTED, 1)
```
smrz2001, line 100: Are these still "legacy" 😅?

smrz2001 (approved):
🚀

src/ancillary/anchor-request-params-parser.ts
stbrody: yeah it's not, it's now back to being the main expected path.  I thought about renaming it but then it will break continuity with the existing metric name.  @gvelez17 curious your thoughts on if this is worth renaming now or not?

gvelez17: hm I don't think i named this, i do not see `cas_server_ctrl_legacy_requested_total` on any of the primary cas dashboards (tho i didn't do a comprehensive search).  I think, since a number of metrics are in flux anyway, I would not worry about historical continuity right now, go ahead and rename it appropriately!

stbrody: renamed to `ctrl_json_requested`, which is a more descriptive and accurate parallel to `ctrl_car_requested`

PR #1233, opened 2024-06-21
feat: Use env variable to configure number of workers

ukstv:
If env variable `MULTIPROCESS_SIZE` (as a decimal number) is set, it takes precedence over nodejs `cpus().length` of `node:os`.
That should fix a bug of "having" 32 processors on k8s regardless of the underlying number of CPUs.

### DIFF
--- a/src/ancillary/multiprocess.ts
+++ b/src/ancillary/multiprocess.ts
@@ -1,6 +1,43 @@
 import EventEmitter from 'node:events'
 import cluster from 'node:cluster'
 import { cpus } from 'node:os'
+import { readFileSync } from 'node:fs'
+
+/**
+ * Read a number from a file.
+ */
+function readNumber(filename: string | URL): number {
+  return Number(readFileSync(filename, 'utf-8'))
+}
+
+/**
+ * Read CPU limits from /sys/fs.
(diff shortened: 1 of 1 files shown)


src/ancillary/multiprocess.ts
```diff
@@ -64,7 +64,10 @@ export class Multiprocess extends EventEmitter {
       this.work()
       return
     }
-    let processes = options.workers || cpus().length // TODO workers = -1 or undef means no workers
```
smrz2001, line 67: Would be good to understand why `cpus()` is returning incorrect results in the first place.

smrz2001: Otherwise this means having to constantly keep the env var updated when updating CAS resources in k8s.

ukstv: https://github.com/nodejs/node/issues/28762

ukstv: And changed the code to use settings from what's inside the Docker.

smrz2001: Nice!! Thanks 😊

smrz2001 (approved):
🚀

PR #1234, opened 2024-07-03
Move auth logic from api gateway lambda to cas codebase

ukstv:
Add auth logic to CAS that permits a request if a DID requestor is in an allowlist.


src/auth/auth.middleware.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    // Use auth lambda
+    const didFromHeader = req.header('did')
+    if (didFromHeader && req.body) {
+      const digest = buildBodyDigest(req.header('Content-Type'), req.body)
+      if (req.header('digest') === digest) {
+        ServiceMetrics.count(METRIC_NAMES.AUTH_ALLOWED, 1, { did: didFromHeader })
+        return next()
+      } else {
+        logger?.verbose(`Disallowed: Auth lambda: Invalid digest`)
```
christianlavoie, line 57: Let's throw in a metric here too, but we probably do NOT want the didFromHeader here (because it's effectively an unbounded size list, which could lead to unbounded memory usage and an external DoS vector), so a simple count please.

src/auth/auth.middleware.ts
```diff
@@ -0,0 +1,145 @@
...
+
+      const body = req.body
+      const contentType = req.header('Content-Type')
+      const digestCalculated = buildBodyDigest(contentType, body)
+      const isCorrectDigest = digestCalculated == digest
+      if (!isCorrectDigest) {
+        logger?.verbose(`Disallowed: Incorrect digest for DID ${did}`)
+        return disallow(res)
+      }
+      ServiceMetrics.count(METRIC_NAMES.AUTH_ALLOWED, 1, { did: did })
```
christianlavoie, line 96: Won't that double count successes?

christianlavoie: I mean, since the same request could be counted at L54

src/auth/auth.middleware.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+    return next()
+  }
+}
+
+function disallow(res: Response): Response {
+  ServiceMetrics.count(METRIC_NAMES.AUTH_DISALLOWED, 1)
```
christianlavoie, line 103: oh, I see. We could also have the reason for disallowing a request as a label, might come in handy debugging at some point in the future.

ukstv: I heard bunch of negative talks about high cardinality. Probably it is alright here, as it is like ~10 reasons for disallowing.

src/auth/auth.middleware.ts
christianlavoie: Also, let's throw in `relaxed` (=0 or =1) as a label here (and just above)  too, it makes some analyses easier.

src/auth/auth.middleware.ts
ukstv: That would not double the counter. L54 is immediately followed by `return next()` terminal, so L96 is not going to be reached.

https://github.com/ceramicnetwork/ceramic-anchor-service/pull/1234/files/9a0ea8636ca8b61d473ccd1126602f1ec72a2bfa#diff-56f006a2ae9c4ee0e62a81ef1d1a0fd7244216124906e7b330a1d3d33231402dR55

christianlavoie (changes requested):
Just looked at the metrics.

ukstv:
I think I addressed your concerns @christianlavoie

src/auth/auth.middleware.ts
christianlavoie: Yeah, for me high cardinality starts at "dozens", plural, maybe hundred or so. Right now it's just a (I think) minor memory issue. And it particular when you have many labels with dozens of values of each -- the combinatorial explosion thing is real.

But I've managed services with a final cardinality in the hundreds without any problem other than somewhat high prometheus memory usage.

christianlavoie (approved):
Good for me, but I didn't look at the code itself, just the monitoring bits. Please wait for @smrz2001 to peek at the code itself please :)

ukstv:
> Good for me, but I didn't look at the code itself, just the monitoring bits. Please wait for @smrz2001 to peek at the code itself please :)

Of course!

src/auth/auth.middleware.ts
```diff
@@ -0,0 +1,155 @@
...
+      if (!isAllowedDID(did, opts)) {
+        logger?.verbose(`Disallowed: ${did}`)
+        return disallow(res, DISALLOW_REASON.DID_ALLOWLIST_REJECTED)
+      }
+
+      const body = req.body
+      const contentType = req.header('Content-Type')
+      const digestCalculated = buildBodyDigest(contentType, body)
+      const isCorrectDigest = digestCalculated == digest
+      if (!isCorrectDigest) {
```
smrz2001, line 101:
```suggestion
      if (digestCalculated !== digest) {
```

src/auth/auth.middleware.ts
smrz2001: Looks like the `disallow` method updates a metric for the disallowed reason. Can add `relaxed` to it as well.

smrz2001 (approved):
I like the code organization! Apart from the couple of minor comments, this looks good to me.

ukstv:
> I like the code organization! Apart from the couple of minor comments, this looks good to me.

"the couple" of comments? I just see one. What did I miss @smrz2001 ? o_O

